PR #11, opened 2020-02-21
Movies list add form

AlexanderShpilka:
https://alexandershpilka.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.js
+++ b/src/components/NewMovie/NewMovie.js
@@ -1,13 +1,109 @@
 import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import './NewMovie.scss';
 
 export class NewMovie extends Component {
-  state = {};
+  state = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
+
+  handleUserInput = (evt) => {
+    const { name } = evt.target;
+    const { value } = evt.target;
+
+    this.setState({ [name]: value });
+  };
+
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,109 @@
...
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
+
+  handleUserInput = (evt) => {
+    const { name } = evt.target;
+    const { value } = evt.target;
```
irenhh: combine this into one line

src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,109 @@
...
+
+    this.setState({ [name]: value });
+  };
+
+  handleFormSubmit = (evt) => {
+    evt.preventDefault();
```
irenhh: event

src/components/NewMovie/NewMovie.js
AlexanderShpilka: done

src/components/NewMovie/NewMovie.js
AlexanderShpilka: done but... :)
Previously I was taught that it's better to avoid using 'event' parameter name 'cause there is self-titled Window property: https://developer.mozilla.org/en-US/docs/Web/API/Window/event

AlexanderShpilka:
> add validations to all fields

Can I skip this requirement as this one is optional according to the task description?

PR #12, opened 2020-02-21
solution without validation

morfeii:
https://morfeii.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.js
+++ b/src/components/NewMovie/NewMovie.js
@@ -1,13 +1,111 @@
 import React, { Component } from 'react';
+import PropTypes from 'prop-types';
 
 export class NewMovie extends Component {
-  state = {};
+  state = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  }
+
+  handleChange = ({ target }) => {
+    const { name, value } = target;
+
+    this.setState({
+      [name]: value,
+    });
+  }
+
+  handleSubmit = (event) => {
+    event.preventDefault();
+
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,111 @@
...
+    });
+  }
+
+  handleSubmit = (event) => {
+    event.preventDefault();
+
+    const { title, description, imgUrl, imdbUrl, imdbId } = this.state;
+    const { addMovie } = this.props;
+
+    addMovie({
```
irenhh: what about an id field for movie

src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,178 @@
...
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    } = this.state;
+
+    // eslint-disable-next-line max-len
+    const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@.\w_]*)#?(?:[.!/\\\w]*))?)$/;
+
+    this.setState((prevState) => {
```
irenhh: to avoid calling setState twice for sure, it would be better if you create a pure function that validates fields and call setState in if else statement where you call setState to set errors if there are any errors according to the validation function output, otherwise you call addTodo and setState to reset all fields

PR #14, opened 2020-02-23
Develop

aleksandrDnieper:
[DEMO LINK](https://aleksandrDnieper.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.js
+++ b/src/components/NewMovie/NewMovie.js
@@ -1,13 +1,222 @@
 import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import './NewMovie.scss';
+
+function isValid(obj) {
+  return Object.values(obj).every(value => value === true);
+}
 
 export class NewMovie extends Component {
-  state = {};
+  state = {
+    movie: {
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    },
+    error: {
+      titleError: false,
+      imgUrlError: false,
+      imdbUrlError: false,
+      imdbIdError: false,
+    },
+    validation: {
+      titleValid: false,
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,222 @@
 import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import './NewMovie.scss';
+
+function isValid(obj) {
+  return Object.values(obj).every(value => value === true);
```
irenhh:
```suggestion
  return Object.values(obj).every(value => value);
```

src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,222 @@
...
+
+    this.setState(prevState => ({
+      movie: {
+        ...prevState.movie,
+        [name]: value.trim(),
+      },
+    }));
+  };
+
+  onBlur = ({ target }) => {
```
irenhh: one setState per method

irenhh: write all code inside one setState

src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,222 @@
...
 
   render() {
+    const {
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    } = this.state;
+    const valid = { ...this.state.validation };
```
irenhh: why do you copy value

src/components/NewMovie/NewMovie.js
aleksandrDnieper: could you give me a link to an article, how should I did this

irenhh: you changed it just like expected, good job!

PR #16, opened 2020-02-24
Develop

delenjer:
[DEMO LINK](https://delenjer.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.js
+++ b/src/components/NewMovie/NewMovie.js
@@ -1,13 +1,164 @@
 import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import './NewMovie.scss';
 
 export class NewMovie extends Component {
-  state = {};
+  state = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    error: false,
+    imdbUrl: '',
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,164 @@
...
+    }
+
+    if (!pattern.test(imdbUrl)) {
+      isError = true;
+
+      this.setState({
+        error: true,
+      });
+    }
+
```
gestap: в тебе один error на 2 поля, виправи це

image

irenhh:
if I corrected error, it should disappear immediately so that I know this for sure + add validations to other fields so that I can't add just spaces

src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,164 @@
...
+  handleChange = ({ target }) => {
+    const { name, value } = target;
+
+    this.setState({
+      [name]: value,
+    });
+  };
+
+  handleSubmit = (event) => {
```
irenhh: one setState per method

PR #18, opened 2020-02-24
add solution

YMagrelo:
[DEMO LINK](https://YMagrelo.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.js
+++ b/src/components/NewMovie/NewMovie.js
@@ -1,13 +1,98 @@
 import React, { Component } from 'react';
+import './NewMovie.scss';
+import PropTypes from 'prop-types';
 
 export class NewMovie extends Component {
-  state = {};
+  state = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
+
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,98 @@
...
+
+    this.setState({
+      [name]: value,
+    });
+  }
+
+  handleSubmit = (event) => {
+    event.preventDefault();
+
+    const { title, description, imgUrl, imdbUrl, imdbId } = this.state;
```
irenhh: place code inside brackets on the new line

src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,170 @@
...
+
+    const {
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    } = this.state;
+
+    if (title.trim() === '') {
```
irenhh: to avoid calling setState twice for sure, it would be better if you create a pure function that validates fields and call setState in if else statement where you call setState to set errors if there are any errors according to the validation function output, otherwise you call addTodo and setState to reset all fields

PR #83, opened 2020-09-08
sol

AlisaVlasova:
[DEMO LINK](https://AlisaVlasova.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.js
+++ b/src/components/NewMovie/NewMovie.js
@@ -1,13 +1,97 @@
 import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import './NewMovie.scss';
 
 export class NewMovie extends Component {
-  state = {};
+  state = {
+    title: '',
+    description: '',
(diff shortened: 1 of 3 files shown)


rmnkk (changes requested):
There is a possibility to add a movie without a title. Because your inputs allow receiving spaces.
https://prnt.sc/udjcj6

src/components/NewMovie/NewMovie.js
```diff
@@ -19,7 +19,7 @@ export class NewMovie extends Component {
 
   handleChange = (event) => {
     this.setState({
-      [event.target.name]: event.target.value,
+      [event.target.name]: event.target.value.trim(),
```
rmnkk, line 22: It's better to use trimStart, then you could add more than one word. Cause now it's possible to add only one word

rmnkk (approved):

Denchepornyuk (approved):

PR #149, opened 2020-11-04
Solution

Feuerbach98:
- [DEMO LINK](https://feurbach98.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -9,8 +9,19 @@ export class App extends Component {
     movies: moviesFromServer,
   };
 
-  addMovie = (movie) => {
-    // put your code here
+  addMovie = (title, description, imgUrl, imdbUrl, imdbId) => {
+    this.setState(state => ({
+      movies: [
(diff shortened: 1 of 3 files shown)


KirillLutsenko (approved):

aliceblunt8 (changes requested):
let's create a separate component for Input and render them through map

aliceblunt8 (changes requested):
1) form should be cleared after the submit (but it isn't)
2) mb show some errors, I couldn't submit a form several times, but I didn't understand why

Feuerbach98:
> 1. form should be cleared after the submit (but it isn't)
> 2. mb show some errors, I couldn't submit a form several times, but I didn't understand why

Fixed cleaning, validation in this task is optional, I'll do it when I figure out the fetch

aliceblunt8 (approved):

aliceblunt8 (approved):
approve

PR #212, opened 2021-01-28
Solution

viktor-morhun:
without css, input component will be added before saturday
- [DEMO LINK](https://viktor-morhun.github.io/react_movies-list-add-form/)


aionova (approved):
Good job!

PR #247, opened 2021-03-31
solution

yyashnyi:
[DEMO LINK](https://yashnyi.github.io/react_movies-list-add-form/)


SofiiaAsieieva (approved):
good job:fire: but I would share the logic of the NewMovie component

mriynyk (approved):

PR #280, opened 2021-05-24
sol

Ihor-Karpyn:
- [DEMO LINK](https://Ihor-Karpyn.github.io/react_movies-list-add-form/)


Burunduk003 (approved):
Very interesting solution and good styling! One logic issue I see here. If you disable the button before the user even tries to enter data, he or she cant see an error message. They didn't know what fields are required, so they can consider your site is broken and the button didn't work in any cases.

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,52 @@
...
+          className={classNames(
+            'textField', { redBorder: isError },
+          )}
+          onChange={event => onChange(event)}
+          onBlur={event => onFocus(event)}
+        />
+      </label>
+      <p
+        className={classNames(
+          'errorMes', { visible: isError },
```
VinnieJ-2k20, line 31: Don't use shortened words in names.

```suggestion
          'errorMessage', { visible: isError },
```

src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,186 @@
...
+      value: '', isError: true, isErrorVisible: false, isRequired: true,
+    },
+    imdbId: {
+      value: '', isError: true, isErrorVisible: false, isRequired: true,
+    },
+  };
+
+  componentDidMount() {
+    this.initialState = this.state;
+  }
```
VinnieJ-2k20, line 28:
```suggestion
  componentDidMount() {
    this.initialState = {
      ...this.state,
    };
  }
```

VinnieJ-2k20 (approved):
Very interesting solution 👍

PR #282, opened 2021-05-27
react_movies-list-add-form

IvanGrekov:
[DEMO LINK](https://ivangrekov.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.scss
+++ b/src/components/NewMovie/NewMovie.scss
@@ -14,17 +14,17 @@ $error-color: red;
     padding: 5px 10px;
     border: 1px solid grey;
 
-    &_textarea {
-      min-height: 4em;
-      resize: vertical;
-    }
-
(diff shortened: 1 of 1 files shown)


rauzok (approved):
Nice!

src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,299 @@
...
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    };
+
+    this.props.onAdd(newMovie);
+
+    return this.setState({
```
VinnieJ-2k20, line 121:
```suggestion
    this.setState({
```

VinnieJ-2k20 (approved):
I'll accept because it works correctly.

If you use `setState` and want to finish execution of the function, its better to call `setState` first and then put `return`.

PR #283, opened 2021-05-28
Solution

Burunduk003:
- [DEMO LINK](https://Burunduk003.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.js
+++ b/src/components/NewMovie/NewMovie.js
@@ -1,13 +1,240 @@
 import React, { Component } from 'react';
+import classNames from 'classnames';
+import './NewMovie.scss';
+import PropTypes from 'prop-types';
 
 export class NewMovie extends Component {
-  state = {};
+  state = {
+    newMovie: {
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    },
+    error: {
+      title: false,
+      description: false,
+      imgUrl: false,
+      imdbUrl: false,
+      imdbId: false,
+    },
+    invalidImgUrl: false,
+    invalidImdbUrl: false,
+  };
+
+  handleError = () => {
+    const { newMovie } = this.state;
+
+    Object.keys(newMovie).forEach(key => this.setState(state => ({
+      error: {
+        ...state.error,
+        [key]: !state.newMovie[key],
+      },
+    })));
+  }
+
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,240 @@
...
+      this.setState({
+        invalidImdbUrl: false,
+      });
+    }
+
+    if (target.name === 'imdbUrl' && !target.value.match(validator)) {
+      this.setState({
+        invalidImdbUrl: true,
+      });
+    }
```
igor-stasiv, line 73: I think you can use the `switch case` construction here, and just set the result of `!target.value.match(validator)` in the state

src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,240 @@
...
+      <form
+        className="NewMovieForm"
+        onSubmit={(event) => {
+          event.preventDefault();
+          this.handleError();
+          if (!Object.values(newMovie).some(value => value === '')) {
+            onAdd(newMovie);
+            this.clearForm();
+          }
+        }}
```
igor-stasiv, line 106: Maybe it will be more clear if you take out of the `render` this function

igor-stasiv (approved):
Good job! But take a look at my comments. Also, the description field did not have to be `required` according to the task condition

src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,235 @@
...
 export class NewMovie extends Component {
-  state = {};
+  state = {
+    newMovie: {
+      title: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    },
+    error: {
```
VinnieJ-2k20, line 14:
```suggestion
    errors: {
```

src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,235 @@
...
+        ...state.newMovie,
+        [key]: '',
+      },
+
+      error: {
+        ...state.error,
+        [key]: false,
+      },
+    })));
+  }
```
VinnieJ-2k20, line 88: No need to use loops here, better to simply use setState with cleared fields.

VinnieJ-2k20 (approved):
Good job 👍 

BTW, according to task guidelines, fields should be validated on blur (when focus is taken away from input).

PR #287, opened 2021-05-28
Solution

TimurHrynenko:
- [DEMO LINK](https://TimurHrynenko.github.io/react_movies-list-add-form/)

Only one question: why regexp form readme always returns false?


VinnieJ-2k20 (approved):
Nice looking solution.
But you are really making it overcomplicated.
For small project like this there's no real need to create custom `Input` component.
Regarding regexp validation - try debugging it and how you pass arguments between functions. Because it should work with valid img links:
image

PR #290, opened 2021-05-30
develop

Andrey203:
[DEMO LINK](https://Andrey203.github.io/react_movies-list-add-form/)


VinnieJ-2k20 (changes requested):
Nice looking solution. But I have couple of comments:

1. Let's not use DOM tools working with forms in React. Use `state` object to store input values.
2. There is no real need to create separate component for `input` fields. It makes your project overcomplicated.
3. If you have errors shown for URL fields, they disappear even if you change `title` input. That's not correct.

VinnieJ-2k20 (approved):
Good job 👍

PR #292, opened 2021-05-30
Solution

igor-stasiv:
[DEMO LINK](https://igor-stasiv.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.jsx
```diff
@@ -0,0 +1,223 @@
...
+            onChange={this.handleChange}
+          />
+        </label>
+        <p
+          className={classNames(
+            'new-movie__error', isErrorMessage.title && 'active',
+          )}
+        >
+          Please, enter a correct title
+        </p>
```
Burunduk003, line 131: in what cases this message will appear?

Burunduk003 (approved):
If you disable the button before the user even tries to enter data, he or she cant see an error message. They didn't know what fields are required, so they can consider your site is broken and the button didn't work in any cases.

src/components/NewMovie/NewMovie.jsx
igor-stasiv: I was thinking that the class name `active` is enough for understanding. I meant that if an error is then the message will be active. But maybe there will be better used just conditional rendering without `classNames`

src/components/NewMovie/NewMovie.jsx
```diff
@@ -0,0 +1,220 @@
...
+      title: false,
+      imgUrl: false,
+      imdbUrl: false,
+      imdbId: false,
+    },
+  };
+
+  componentDidMount() {
+    this.defaultState = this.state;
+  }
```
aliceblunt8, line 27: hmm why do you need this?

src/components/NewMovie/NewMovie.jsx
```diff
@@ -0,0 +1,220 @@
...
+        return;
+      case 'imgUrl':
+      case 'imdbUrl':
+        this.setState(state => ({
+          [name]: value,
+          isError: {
+            ...state.isError,
+            [name]: !urlValidation.test(value),
+          },
+        }));
```
aliceblunt8, line 52: leave only cases with imgUrl and imdbUrl here, all other cases should be processed in default case

src/components/NewMovie/NewMovie.jsx
```diff
@@ -0,0 +1,220 @@
...
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    } = this.state;
+    const isError = {
+      title: !title.trim(),
+      imgUrl: !urlValidation.test(imgUrl),
+      imdbUrl: !urlValidation.test(imdbUrl),
+      imdbId: !imdbId.trim(),
+    };
```
aliceblunt8, line 76:
```suggestion
    const { onAdd } = this.props;
    
    const {
      title,
      description,
      imgUrl,
      imdbUrl,
      imdbId,
    } = this.state;
    
    const isError = {
      title: !title.trim(),
      imgUrl: !urlValidation.test(imgUrl),
      imdbUrl: !urlValidation.test(imdbUrl),
      imdbId: !imdbId.trim(),
    };
```

aliceblunt8 (changes requested):

src/components/NewMovie/NewMovie.jsx
igor-stasiv: I use it for reset form fields

src/components/NewMovie/NewMovie.jsx
igor-stasiv: I can't do it, because I don't need to check errors of the `description` field, so `description` is default case

src/components/NewMovie/NewMovie.jsx
aliceblunt8: you can do event.target.reset() inside handleSubmit function instead of this

aliceblunt8 (changes requested):
I left comments

src/components/NewMovie/NewMovie.jsx
igor-stasiv: it doesn't work

aliceblunt8 (approved):

PR #317, opened 2021-05-31
React Movies List Add Form

rauzok:
[DEMO LINK](https://rauzok.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.js
+++ b/src/components/NewMovie/NewMovie.js
@@ -1,13 +1,113 @@
 import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+import './NewMovie.scss';
 
 export class NewMovie extends Component {
-  state = {};
+  state = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
+
+  addEvent = (event, name) => this.setState({ [name]: event.target.value });
+
+  movieSubmit = (event) => {
+    event.preventDefault();
+    this.props.onAdd(this.state);
+
+    this.setState({
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
+  }
 
   render() {
+    const { title, description, imdbId, imdbUrl, imgUrl } = this.state;
+
     return (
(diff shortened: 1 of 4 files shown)


IvanGrekov (approved):
OK, but I think error mustn't be shown at once, how we enter to the page

src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,115 @@
...
-  state = {};
+  state = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
+
+  addEvent = (event) => {
```
VinnieJ-2k20, line 15: Let's use meaningful names. 
```suggestion
  handleChange = (event) => {
```

src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,115 @@
...
+    this.setState({ [event.target.name]: event.target.value });
+  };
+
+  movieSubmit = (event) => {
+    event.preventDefault();
+    this.props.onAdd(this.state);
```
VinnieJ-2k20, line 21: Better option would be to craft new object for adding to the list.
State potentially may have more fields.

src/components/NewMovie/NewMovie.scss
```diff
@@ -0,0 +1,17 @@
...
+  &__text, &__button {
+    display: block;
+    height: 30px;
+  }
+
+  &__error {
+    font-size: 12px;
+    color: red;
+  }
+}
```
VinnieJ-2k20, line 17: Don't forget about extra lines at the end of the files.

VinnieJ-2k20 (changes requested):
Almost done.
I added some comments on code optimisation.
Also, it would be great if error messages would not be visible on page load.

VinnieJ-2k20 (changes requested):
You fixed the comments. Good!
Let's now fixed logic a little bit.

Again - you show red error messages on page load. But the form doesn't have any validations, so it can be submitted with all empty fields.
The simplest step is to add `required` attribute to inputs to prevent that.

mykola-yarchuk (approved):

PR #318, opened 2021-05-31
solution

den-bulaev:
[DEMO LINK](https://den-bulaev.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.jsx
```diff
@@ -0,0 +1,100 @@
...
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
+
+  handleChange = (event) => {
+    this.setState({
+      [event.target.name]: event.target.value,
```
dmytro-abram, line 15: Можна було зробити деструктуризацію

dmytro-abram (approved):
GJ!

src/components/NewMovie/NewMovie.jsx
```diff
@@ -0,0 +1,100 @@
...
+            });
+
+            this.setState({
+              title: '',
+              description: '',
+              imdbId: '',
+              imdbUrl: '',
+              imgUrl: '',
+            });
+          }
```
VinnieJ-2k20, line 88: Film should be added to the list correctly not only when you click submit button, but if you press Enter.
Hint - move logic of adding movie to onSubmit attribute of the form.

VinnieJ-2k20, line 88: Film should be added to the list correctly not only when you click submit button, but if you press Enter.
Hint - move logic of adding movie to onSubmit attribute of the form.

VinnieJ-2k20 (changes requested):
Almost done.

Fix submit logic.

VinnieJ-2k20 (approved):
I'll approve because you completed all basic task requirements.
But it would be great to add at least some validations to prevent submitting of empty form.
The easiest way is to add `required` attribute to form inputs.

PR #319, opened 2021-05-31
solution

dmytro-abram:
[DEMO LINK](https://dmytro-abram.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.js
+++ b/src/components/NewMovie/NewMovie.js
@@ -1,13 +1,100 @@
 import React, { Component } from 'react';
+import PropTypes from 'prop-types';
 
 export class NewMovie extends Component {
-  state = {};
+  state = {
+    newMovie: {
+      title: '',
+      description: '',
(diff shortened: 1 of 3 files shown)


den-bulaev (approved):
Good work!

src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,110 @@
...
 export class NewMovie extends Component {
-  state = {};
+  state = {
+    newMovie: {
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    },
```
VinnieJ-2k20, line 12: Better to store values as separate fields and craft an object only on form submit.

VinnieJ-2k20 (approved):
Good job 👍 
Made one small comment.
Also, I would add `required` to inputs where necessary to avoid submitting an empty form.

PR #326, opened 2021-05-31
Solution

bohdanklius:
- [DEMO LINK](https://bohdanklius.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.js
+++ b/src/components/NewMovie/NewMovie.js
@@ -1,13 +1,97 @@
-import React, { Component } from 'react';
+import React from 'react';
+import PropTypes from 'prop-types';
+import './NewMovie.scss';
 
-export class NewMovie extends Component {
-  state = {};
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.js
```diff
@@ -1,13 +1,97 @@
...
+
+    this.setState({
+      [name]: value,
+    });
+  }
+
+  handleSubmit = (event) => {
+    event.preventDefault();
+
+    this.props.onAdd(this.state);
```
VinnieJ-2k20, line 25: Don't pass the whole state object. Craft a separate object for `newMovie` here and use it.

VinnieJ-2k20 (approved):
Good job 👍 
I added some comment on increasing the reliability of your code.
Also, it would be great to prevent submitting of empty form.
Easiest way to do it - is to add `required` attribute to inputs.

PR #397, opened 2021-09-10
solution without validation

VeronikaLushchik:
[DEMO LINK](https://VeronikaLushchik.github.io/react_movies-list-add-form/)


SashaRudenko (approved):

Elena-brv (changes requested):
errors should not be displayed when the page has just loaded
image

Elena-brv (approved):

PR #399, opened 2021-09-13
The task has been done but without url validate

Moroz-Dmytro:
[DEMO LINK](https://Moroz-Dmytro.github.io/react_movies-list-add-form/)


nastyyvolk (approved):
Well done 🔥

Elena-brv (changes requested):
errors should not be displayed when the page has just loaded
image

Elena-brv (approved):
great! interesting solution

PR #401, opened 2021-09-14
react movies list add form

DmytroMykoliv:
[DEMO LINK](https://dmytromykoliv.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,16 +1,157 @@
-import { Component } from 'react';
+import React from 'react';
 
-type Props = {};
-type State = {};
+type Props = {
+  onAddMovie: (movie: Movie) => void,
+};
 
-export class NewMovie extends Component<Props, State> {
-  state: State = {};
+type State = {
+  title: string,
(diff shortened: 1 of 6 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,142 @@
...
+      <>
+        <h1 className="App__new-movie">Add a Movie</h1>
+        <form className="needs-validation" onSubmit={this.getNewMovie}>
+          <div className="mb-3">
+            <input
+              type="text"
+              name="title"
+              className="form-control"
+              placeholder="Please enter a title"
+              value={this.state.title}
```
sasha-morozov, line 75: I would have used destructuring once instead of writing "this.state..." multiple times

sasha-morozov (approved):
Good job!)

src/components/NewMovie/NewMovie.tsx
DmytroMykoliv: Yes, you're right )

danheim (approved):

PR #402, opened 2021-09-14
solution

dima-tkachenko:
[DEMO LINK](https://dima-tkachenko.github.io/react_movies-list-add-form/)


sasha-morozov (approved):
Good job!
However, I would suggest creating a separate component for Form and Input.

danheim (approved):

PR #432, opened 2021-09-15
simple solution done

SashaLifashkin:


VVHimiak (approved):
Nice work! But I recommend using CSS frameworks instead of doing extra work, but it is optional because you approve your skill in CSS!)

danheim (approved):

PR #433, opened 2021-09-15
solution

Vova-Stoliar:
[DEMO LINK](https://Vova-Stoliar.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -11,15 +11,17 @@ type State = {
 
 const inputItems = ['title', 'description', 'imgUrl', 'imdbUrl', 'imdbId'];
 
+const movieBody = {
+  title: '',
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,62 @@
...
+      return {
+        movie: {
+          ...movie,
+          [name as string]: value,
+        },
+      };
+    });
+  };
+
+  handleSubmit = (event: React.FormEvent<EventTarget>) => {
```
MaksymBeliever, line 40: Just `React.FormEvent` instead of `React.FormEvent<EventTarget>`

MaksymBeliever (changes requested):
The form is not cleaned after submitting

Elena-brv (approved):

PR #500, opened 2021-11-17
working but without styles

denys-anosov:
[DEMO LINK](https://denys-anosov.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,15 +1,123 @@
...
+          this.props.onAdd(this.state);
+
+          this.setState({
+            title: '',
+            description: '',
+            imgUrl: '',
+            imdbUrl: '',
+            imdbId: '',
+          });
+        }}
```
dima-narozhnyi, line 52: In this method, a new handler function will be created on each re-render, which is probably not the desired behavior. Сreate a separate method and move this logic there
http://joxi.ru/8AnYeojtNkM312

dima-narozhnyi (changes requested):
Looks like hell! Reject))
😈

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,15 +1,138 @@
...
+
+  inputHandler = (event: InputEvent) => {
+    const { name, value } = event.target;
+
+    this.setState((prevState) => (
+      {
+        ...prevState,
+        [name]: value,
+      }
+    ));
```
VinnieJ-2k20, line 35:
```suggestion
    this.setState((prevState) => ({
      ...prevState,
      [name]: value,
    }));
```

VinnieJ-2k20 (approved):
Good job 👍

PR #515, opened 2021-11-28
Create a form for adding a film to the list

dmtrhrytsak:
[MOVIES](https://dmtrhrytsak.github.io/react_movies-list-add-form/)


aionova (approved):
You added nice validation for urls. It would be great if submit button was disabled if one of the inputs has invalid data

dmtrhrytsak:
> You added nice validation for urls. It would be great if submit button was disabled if one of the inputs has invalid data

It is disabled)

PR #522, opened 2021-12-15
Solution

Vitaliy-0:
[demo](https://vitaliy-0.github.io/react_movies-list-add-form/)


ImIkator (approved):
Well done! Everything like on a task - great

AlinaMatuschak (changes requested):
1 imgUrl, imdbUrl - should be valid urls
2 use trim for check inputs value

AlinaMatuschak (approved):
🚀

PR #525, opened 2021-12-16
solution

skachkov23:
[DEMO](https://skachkov23.github.io/react_movies-list-add-form/)


volodymyr-soltys97 (approved):

AlinaMatuschak (changes requested):

AlinaMatuschak (approved):

PR #526, opened 2021-12-16
Solution

volodymyr-soltys97:
[DEMO LINK](https://volodymyr-soltys97.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,16 +1,109 @@
-import { Component } from 'react';
+import React, { Component } from 'react';
 
-type Props = {};
-type State = {};
+type Props = {
+  addMovie: (movies: Movie) => void
+};
+
+type State = {
+  title: string,
+  description: string,
+  imgUrl: string,
+  imdbUrl: string,
+  imdbId: string
+};
 
 export class NewMovie extends Component<Props, State> {
-  state: State = {};
+  state: State = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
+
+  handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    this.setState({
+      title: '',
+      description: '',
+      imgUrl: '',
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,108 @@
...
+      imdbId: '',
+    });
+
+    const {
+      title, description, imdbUrl, imgUrl, imdbId,
+    } = this.state;
+
+    const newMovie = {
+      title, description, imdbUrl, imgUrl, imdbId,
+    };
```
alex-kostomarov, line 50: instead of it, you can use newMovie object right in the state, and puts new values to the object in the state

alex-kostomarov (approved):
Good job!

AlinaMatuschak (changes requested):
6 title, imgUrl, imdbUrl, imdbId should be required

volodymyr-soltys97:
It was an additional task. Isn't it?

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,112 @@
...
+    imdbUrl: '',
+    imdbId: '',
+  };
+
+  changeHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { name, value } = event.target;
+
+    this.setState(state => ({
+      ...state,
+      [name]: value,
```
AlinaMatuschak, line 29: 🔥

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,112 @@
...
+
+    const {
+      title, description, imdbUrl, imgUrl, imdbId,
+    } = this.state;
+
+    const newMovie = {
+      title, description, imdbUrl, imgUrl, imdbId,
+    };
+
+    this.props.addMovie(newMovie);
```
AlinaMatuschak, line 52: Make sure that you can’t add movie with all empty values

AlinaMatuschak (changes requested):
fix @alex-kostomarov comment

AlinaMatuschak (approved):

PR #527, opened 2021-12-16
completed react movies list add form

i-stanko:


msaikun (approved):

AlinaMatuschak (changes requested):
6 title, imgUrl, imdbUrl, imdbId should be required

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,15 +1,123 @@
...
+    imdbUrl: '',
+    imdbId: '',
+  };
+
+  changeHandler = (event:React.ChangeEvent<HTMLInputElement>) => {
+    const { name, value } = event.target;
+
+    this.setState(state => ({
+      ...state,
+      [name]: value,
```
AlinaMatuschak, line 25: 🔥

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,15 +1,123 @@
...
+    }));
+  };
+
+  submitHandler = (event:React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    this.props.onAdd(this.state);
```
AlinaMatuschak, line 31: Make sure that you can’t create movie with all empty values

AlinaMatuschak (changes requested):

AlinaMatuschak (approved):
Users can create movies with spaces without letters 🥲

PR #528, opened 2021-12-16
Solution

lera-permiakova:
[DEMO LINK](https://lera-permiakova.github.io/react_movies-list-add-form/)


AlinaMatuschak (changes requested):
Users can create movies with spaces without letters 🥲

AlinaMatuschak (approved):
Good work 💪. Don't forget to redeploy

PR #554, opened 2021-12-21
init

junglq:


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,15 +1,124 @@
...
+  };
+
+  handleChange = (event: React.ChangeEvent<HTMLInputElement> |
+  React.ChangeEvent<HTMLTextAreaElement>) => {
+    const { name, value } = event.target;
+
+    this.setState(state => ({
+      newMovie: {
+        ...state.newMovie,
+        [name]: value,
```
AlinaMatuschak, line 30: save trimmed value

AlinaMatuschak (changes requested):
Users can create movies with all empty values 🥲

AlinaMatuschak (approved):

PR #568, opened 2022-01-26
solution

samson-sim:


Denchepornyuk (commented):
Screenshot 2022-02-01 at 19 08 09
inputs should be required

Denchepornyuk (changes requested):
inputs should be required

Denchepornyuk (approved):

PR #625, opened 2022-03-15
Solution

ArtemPaskall:
[DEMO LINK](https://ArtemPaskall.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,137 @@
...
+            imdbUrl,
+            imdbId,
+          });
+          setTitle('');
+          setDescription('');
+          setImgUrl('');
+          setImdbUrl('');
+          setImdbId('');
+        }
+      }}
```
Elena-brv, line 46: create a separate method for submitting form
also, all this code should be divided into different functions (validate, resetForm etc.)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,137 @@
...
+    >
+      <label
+        htmlFor="title"
+      >
+        Title
+        <br />
+        <input
+          type="text"
+          id="title"
+          className={hasTitleError ? 'error' : ''}
```
Elena-brv, line 56: use classnames library https://www.npmjs.com/package/classnames

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,137 @@
...
+          placeholder="Enter title"
+          value={title}
+          onChange={(event) => {
+            setTitle(event.target.value);
+          }}
+        />
+      </label>
+      <br />
+      <br />
```
Elena-brv, line 65: don't use it, use css for making indents instead (everywhere)

Elena-brv (changes requested):
- the errors should be hidden when smth changes in input 
image

- a description is an optional field, so user should have ability to add a movie without filled-in description

Elena-brv (approved):

PR #633, opened 2022-03-17
React: Movies list - Add Form

yuran007:
[DEMO LINK](https://yuran007.github.io/react_movies-list-add-form/)


aliceblunt8 (changes requested):
I can add an empty movie

Screenshot 2022-03-18 at 14 54 57

PR #710, opened 2022-05-19
done

Yurii837:


AlinaMatuschak (changes requested):
- I can add film when title and imd id just with spaces
Screenshot 2022-05-24 at 13 02 35

- when you add the first film next films you can add without any values 
Screenshot 2022-05-24 at 13 04 37

AlinaMatuschak (changes requested):
спробуй добавити один фільм, усі наступні можна добавляти навіть якщо зовсім нічого не ввів (кнопка "add new film" активна)
Screenshot 2022-05-24 at 20 43 20
Screenshot 2022-05-24 at 20 43 34

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,219 @@
...
+    setErrors({
+      ...errors,
+      imdbId: errorMessage,
+      description: errorMessage,
+    });
+
+    return errorMessage;
+  };
+
+  const isValid: ()=> boolean = () => {
```
AlinaMatuschak, line 108: try use useMemo (isValid will be boolean value) and add on array dependencies every error:
[
  errors.title,
  errors.imgUrl,
  ...
]

or create a separate state for each error and use these states on array dependencies
In this way, you won’t need currentValid at all and you will delete useEffect without array dependencies

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,219 @@
...
+            autoComplete="off"
+          />
+          {errors.imdbId.length > 0
+            && <p className="form__error-message">{errors.imdbId}</p>}
+        </div>
+
+        <button
+          type="submit"
+          className="form__button"
+          disabled={!currentValid}
```
AlinaMatuschak, line 212: did you try to use isValid here?

AlinaMatuschak (approved):
you wrote hard logic (you can make it easier). Will be great if you try to check my comments

PR #727, opened 2022-06-02
Develop

Bohdan-Dubravin:
https://Bohdan-Dubravin.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,16 +1,187 @@
-import { Component } from 'react';
+import { useEffect, useState } from 'react';
+import './NewMovie.scss';
 
-type Props = {};
-type State = {};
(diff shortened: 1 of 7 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,187 @@
...
+        ...prev,
+        [name]: value,
+      }
+    ));
+  }
+
+  function handleSubmit(e: React.FormEvent<HTMLFormElement>) {
+    e.preventDefault();
+
+    onAdd(newMovie);
```
aliceblunt8, line 49: also clear the form after submit, cause I can add duplicate films (also you need to add check for the same title before adding the film)
Screenshot 2022-06-03 at 12 52 01

aliceblunt8 (changes requested):

PR #730, opened 2022-06-10
movie list

Andriyk0:
[DEMO LINK](https://Andriyk0.github.io/react_movies-list-add-form/)


SavMary (commented):
Good!

SavMary (approved):
Good! Some advice from me: Borders should be red only when there is an error. The description field may not be required

Denchepornyuk (approved):

PR #734, opened 2022-06-11
solution v1

alexandra-protyanova:


vlasiuk-anatolii (approved):
Film adds with wrong data. Did not implement opportunity check right URL

Anna-Tkachenko (approved):
Good job!

PR #736, opened 2022-06-11
solutionI

EnjSadman:
[DEMO LINK](https://EnjSadman.github.io/react_movies-list-add-form/)


vlasiuk-anatolii (approved):
You did't implement handle empty fields. They are empty and film adds.

Anna-Tkachenko (approved):
As user I can't understand why I can't submit the form with empty fields and what you expect to receive in url fields
Will be better to show corresponding errors

PR #737, opened 2022-06-11
solution not 100%

ZUBOV-ILLIA:
[DEMO LINK](https://ZUBOV-ILLIA.github.io/react_movies-list-add-form/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Movies list - Add Form
 - Replace `<your_account>` with your Github username in the
-  [DEMO LINK](https://<your_account>.github.io/react_movies-list-add-form/)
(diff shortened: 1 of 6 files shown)


vlasiuk-anatolii (approved):
When passed wrong data the button should be disabled. Idea add test information is cool.

Anna-Tkachenko (changes requested):
I can't submit form with wrong data but I don't see any errors 🥲
image

Anna-Tkachenko (changes requested):
I can't add movie🥲
image

Anna-Tkachenko (approved):
Good job!
Just will be better to show errors near the input instead of as a placeholder
Whey user enter some text - he doesn't see error text

PR #743, opened 2022-06-13
react_movies-list-add-form

anastasiia-nikita:
[DEMO LINK](https://anastasiia-nikita.github.io/react_movies-list-add-form/)


Anna-Tkachenko (approved):
Good job!
Just as for user it's difficult to understand why submit button is initially disabled, I don't see any errors or messages

PR #745, opened 2022-06-13
movie form

kitaminary:
[DEMO LINK](https://kitaminary.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,16 +1,216 @@
-import { Component } from 'react';
+/* eslint-disable max-len */
+import classNames from 'classnames';
+import React, { useEffect, useState } from 'react';
 
-type Props = {};
-type State = {};
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
 
-export class NewMovie extends Component<Props, State> {
-  state: State = {};
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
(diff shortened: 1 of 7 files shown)


v-vytrykush (commented):
Great job!

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,216 @@
...
+    } else {
+      setUrlImgError(false);
+    }
+  };
+
+  const validateUrl = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (!validUrl.test(event.target.value)) {
+      setUrlError(true);
+    } else {
+      setUrlError(false);
```
Anna-Tkachenko, line 58: It's redundant here, you call this validate function only in submit handler, and it should set true if there is some error in the form
But if no - you shouldn't set this value to the false, you did it after adding new movie

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,216 @@
...
+    } else {
+      setDisabledButton(true);
+    }
+  }, [imdbUrl, imgUrl, title, description, imdbId]);
+
+  const validateImageUrl = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (!validUrl.test(event.target.value)) {
+      setUrlImgError(true);
+    } else {
+      setUrlImgError(false);
```
Anna-Tkachenko, line 50: same here

Anna-Tkachenko (approved):
As for user it's difficult to understand why submit button is initially disables, I don't see any errors or messages

PR #762, opened 2022-06-22
solution, advanced tasks not done

efdalyalcin:
- [DEMO LINK](https://efdalyalcin.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -25,23 +25,23 @@ export const NewMovie: FC<Props> = ({ addMovie }) => {
 
   const validateInput = () => {
     if (!newMovie.title) {
-      setErrors({ ...errors, hasTitleError: true });
+      setErrors(() => ({ ...errors, hasTitleError: true }));
     }
 
     if (!newMovie.description) {
-      setErrors({ ...errors, hasDescrError: true });
+      setErrors(() => ({ ...errors, hasDescrError: true }));
     }
 
     if (!newMovie.imgUrl) {
-      setErrors({ ...errors, hasImgUrlError: true });
+      setErrors(() => ({ ...errors, hasImgUrlError: true }));
     }
 
     if (!newMovie.imdbUrl) {
-      setErrors({ ...errors, hasImdbUrlError: true });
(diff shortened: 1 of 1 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,163 @@
...
+  const [hasDescrError, setHasDescrError] = useState(false);
+
+  const [imgUrl, setImgUrl] = useState('');
+  const [hasImgUrlError, setHasImgUrlError] = useState(false);
+
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [hasImdbUrlError, setHasImdbUrlError] = useState(false);
+
+  const [imdbId, setImdbId] = useState('');
+  const [hasImdbIdError, setHasImdbIdError] = useState(false);
```
ihor-jpeg, line 23: why not simply create two state variables, but make them objects? one object that stores values for the inputs and one for errors?

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,163 @@
...
+
+  const movie = {
+    title,
+    description,
+    imgUrl,
+    imdbUrl,
+    imdbId,
+  };
```
ihor-jpeg, line 31: why do you need to create it here? is it used before we submit the form?

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,163 @@
...
+      && imgUrl
+      && imdbUrl
+      && imdbId) {
+      addMovie(movie);
+      setTitle('');
+      setDescription('');
+      setImgUrl('');
+      setImdbUrl('');
+      setImdbId('');
+    }
```
ihor-jpeg, line 71: make it look better :)

ihor-jpeg (changes requested):

src/components/NewMovie/NewMovie.tsx
efdalyalcin: I created 2 objects and stored the info in there, but now I am thinking there are lots of object copying, isn't it a bad thing for performance? 

like this `setErrors({ ...errors, hasTitleError: true });`

PR #766, opened 2022-06-23
task completed

andreisucman:
[DEMO LINK](https://andreisucman.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -11,7 +11,6 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   const [hasTitleError, setHasTitleError] = useState(false);
 
   const [description, setDescription] = useState('');
-  const [hasDescriptionError, setHasDescriptionError] = useState(false);
 
   const [imgUrl, setImgUrl] = useState('');
   const [hasImgUrlError, setHasImgUrlError] = useState(false);
@@ -33,10 +32,6 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   const validateInput = () => {
     if (!title) {
       setHasTitleError(true);
-    }
-
-    if (!description) {
-      setHasDescriptionError(true);
     }
 
     if (!imgUrl) {
@@ -88,12 +83,11 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
         name="description"
         id=""
         placeholder="Enter description"
-        className={cn('textarea', 'my-3', { error: hasDescriptionError })}
+        className={cn('textarea', 'my-3')}
         value={description}
         data-cy="form-description"
         onChange={(event) => {
           setDescription(event.target.value);
-          setHasDescriptionError(false);
         }}
       />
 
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,145 @@
...
-  render() {
-    return (
-      <form>
-        Put the form here
-      </form>
-    );
-  }
-}
+  const [description, setDescription] = useState('');
+  const [hasDescriptionError, setHasDescriptionError] = useState(false);
```
MabuzaM, line 14: I don't think you need to validate description field since it's not required.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,145 @@
...
+
+  return (
+    <form
+      className="form"
+      onSubmit={submitValidForm}
+    >
+      <input
+        type="text"
+        placeholder="Enter movie title"
+        className={cn('input', { error: hasTitleError })}
```
MabuzaM, line 78: You are validating inputs. I think you forgot to add styles to your css file that's why it's not working

MabuzaM (changes requested):
Good work. Just a few suggestions below.

src/components/NewMovie/NewMovie.tsx
andreisucman: True, removed.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,139 @@
...
+  const [description, setDescription] = useState('');
+
+  const [imgUrl, setImgUrl] = useState('');
+  const [hasImgUrlError, setHasImgUrlError] = useState(false);
+
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [hasImdbUrlError, setHasImdbUrlError] = useState(false);
+
+  const [imdbId, setImdbId] = useState('');
+  const [hasImdbIdError, setHasImdbIdError] = useState(false);
```
ihor-jpeg, line 22: dont forget that you can also use objects as a state, instead of creating multiple primitive state values you can create one and make it an object for those properties that relate to something in common (form inputs)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,139 @@
...
+
+  const newMovie = {
+    title,
+    description,
+    imgUrl,
+    imdbUrl,
+    imdbId,
+  };
```
ihor-jpeg, line 30: why do we need to create it here?
this only needs to be created when we submit the form

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,139 @@
...
+
+  const validateInput = () => {
```
ihor-jpeg, line 32: useCallback

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,139 @@
...
+    if (!imdbUrl) {
+      setHasImdbUrlError(true);
+    }
+
+    if (!imdbId) {
+      setHasImdbIdError(true);
+    }
+  };
+
+  const submitValidForm = (event: React.FormEvent<HTMLFormElement>) => {
```
ihor-jpeg, line 50: useCallback

ihor-jpeg (changes requested):

PR #776, opened 2022-06-28
initial

Sardorbek1122:
[DEMO LINK](https://Sardorbek1122.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,16 +1,106 @@
-import { Component } from 'react';
+import { useState } from 'react';
+import './NewMovie.scss';
 
-type Props = {};
-type State = {};
-
-export class NewMovie extends Component<Props, State> {
-  state: State = {};
-
-  render() {
-    return (
-      <form>
-        Put the form here
-      </form>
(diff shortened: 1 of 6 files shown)


meforoma (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -54,43 +49,55 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
...
-        value={imdbId}
+        value={emptyMovieInputs.imdbId}
         className="form__imbId"
         data-cy="form-imdbId"
         onChange={(event) => {
-          setImdbId(event.target.value);
+          setNewMovie({
+            ...newMovie,
+            imdbId: event.target.value
+          });
```
meforoma, line 100: love this way better that previous, yet each onChange is same here - you could introduce a generic `handleChange` that would use `event.target.name`, like so:
```
const handleChange = (event) => {
  setNewMovie({
            ...newMovie,
            [name]: value
          });
}
```

and you use across all the input fields instead of customising it for each field `onChange={(event) => handleChange(event)}`

PR #783, opened 2022-07-15
Solution

mira1zu:
- [DEMO LINK](https://mira1zu.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,16 +1,227 @@
...
+      }
+
+      default: {
+        setInvalidInputs(
+          invalidInputs.filter(inputName => inputName !== name),
+        );
+        newValidInputs = [...newValidInputs, name];
+        break;
+      }
+    }
```
jstmpelowycz, line 51: I'd rather use simple if/else here

jstmpelowycz (approved):

PR #847, opened 2022-08-03
solution

FANTAZER-NURE:
https://FANTAZER-NURE.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,119 @@
-import { useState } from 'react';
+import { FormEvent, useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (newValue: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [image, setImage] = useState('');
+  const [imbd, setImbd] = useState('');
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,119 @@
...
+        imdbUrl: imbd,
+        imdbId: imbdId,
+      };
+
+      onAdd(movie);
+      setTitle('');
+      setDescription('');
+      setImage('');
+      setImbd('');
+      setImbdId('');
```
Bohdan-Maksymiuk, line 67: pass it to separate callback

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,119 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [image, setImage] = useState('');
+  const [imbd, setImbd] = useState('');
+  const [imbdId, setImbdId] = useState('');
```
Bohdan-Maksymiuk, line 17: also you may try to store these in one state-object

Bohdan-Maksymiuk (changes requested):

src/components/NewMovie/NewMovie.tsx
FANTAZER-NURE: tried that and figured out, that its uncomfortable to use state from object, so decided to leave with separate variables

Bohdan-Maksymiuk (approved):

src/components/NewMovie/NewMovie.tsx
Bohdan-Maksymiuk: You may join q&a session or ask in chat - and I will show you advantages

PR #861, opened 2022-08-16
Solution

Lentyaeev:
[DEMO LINK](https://lentyaeev.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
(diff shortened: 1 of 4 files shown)


Kras4iy (approved):
image

When I try to paste url and after that i switch to another field i recieve a required message. But with this required message i can add a movie

image

src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie
+          onAdd={(movie: Movie) => {
+            setMovies(prevState => [...prevState, movie]);
+          }}
```
solaryasha, line 20: Avoid using inline event handlers

solaryasha (changes requested):
Let's fix Oleksandr comments

VETALEK (approved):
👍

PR #867, opened 2022-08-16
Solution

Sergii2705:
- [DEMO LINK](https://Sergii2705.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -55,7 +55,7 @@ export const TextField: React.FC<Props> = ({
       {hasError && (
         <p className="help is-danger">{`${label} is required`}</p>
       )}
-      {touched && hasErrorPattern && (
+      {touched && !hasError && hasErrorPattern && (
         <p className="help is-danger">
           {`"Error, Invalid URL, path in "${label}" field`}
         </p>


Kras4iy (approved):
image
In your code i have the opportunity to enter correct and incorrect link, maybe it will better, when i enter incorrect link, I will receive another notification like: `incorrect link` or `please enter correct link` . But it was additional condition in this task))). 🚀

src/components/TextField/TextField.tsx
```diff
@@ -19,13 +20,17 @@ export const TextField: React.FC<Props> = ({
...
   const [id] = useState(() => `${name}-${getRandomDigits()}`);
 
   // To show errors only if the field was touched (onBlur)
-  const [touched, setToched] = useState(false);
-  const hasError = touched && required && !value;
+  const [touched, setTouched] = useState(false);
+  const hasError = touched && required
+    && (pattern
+      ? !value.match(pattern)
+      : !value);
```
solaryasha, line 33: I will render an another error message if we don't match RegExp pattern

solaryasha (changes requested):

src/components/TextField/TextField.tsx
Sergii2705: I agree with you guys, I just didn't have time to do it yesterday

VETALEK (approved):
👍

PR #880, opened 2022-08-17
Solution

budanov20:
[DEMO LINK](https://budanov20.github.io/react_movies-list-add-form/)


yaroslav-lizogub (changes requested):
1. NewMovie should check if title, imgUrl, imdbUrl, imdbId are entered when an input looses focus (onBlur) and show an error and a red border if needed (now this work only for title)
2. Disable the submit button until all the required fields are filled;

PR #908, opened 2022-09-05
solution

victorgri:
https://victorgri.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -62,6 +62,7 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
         label="Description"
         value={description}
         onChange={setDescription}
+        required
       />
 
       <TextField
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -43,9 +43,7 @@ export const TextField: React.FC<Props> = ({
           type="text"
           placeholder={`Enter ${label}`}
           value={value}
-          onChange={event => onChange(event
-            .target
-            .value || 'This field is reqired')}
+          onChange={event => onChange(event.target.value)}
           onBlur={() => setToched(true)}
           name="input"
         />


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,111 @@
...
+        };
+
+        onAdd(newMovie);
+
+        setTitle('');
+        setDescription('');
+        setImageUrl('');
+        setImdbUrl('');
+        setImdbID('');
+      }}
```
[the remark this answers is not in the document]
victorgri: Можеш пояснити як це зробити?

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,111 @@
...
+  imdbId: string,
+}
+
+type Props = {
+  onAdd: (newMovie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
```
tetiana-ivanchenko, line 18: remove comments

src/components/NewMovie/NewMovie.tsx
tetiana-ivanchenko, line 52: if you have any questions about how to fixed smth - ask it in the fe_chat, instead of re-requesting review.
you should create separate function with your logic before return, and here just passed this function

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,111 @@
...
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(value) => {
+          setTitle(value);
+        }}
```
tetiana-ivanchenko, line 62:
```suggestion
        onChange={setTitle}
```

tetiana-ivanchenko: fix all similar cases

tetiana-ivanchenko (changes requested):

src/components/NewMovie/NewMovie.tsx
tetiana-ivanchenko: also now I can add film with just spaces in all fields

tetiana-ivanchenko (changes requested):

VitaliyBodnar (changes requested):
- there is no validation for empty text
Screenshot 2023-10-04 at 13 02 54

lerastarynets (changes requested):
Errors should look like so, now you just add error message as inputs' value which is incorrect
Screenshot at Oct 04 14-51-16

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,92 @@
...
+
+    onAdd({
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    });
+
+    setCount(count + 1);
```
lerastarynets, line 39: pls use [prevstate](https://blog.greenroots.info/react-hook-usestate-lazy-initialization-previous-state#heading-how-to-get-the-previous-state-data-and-use-it) for such cases

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #912, opened 2022-09-12
solve the reactMoviesListAddForm problem

tsybulko-natalya:
https://tsybulko-natalya.github.io/react_movies-list-add-form/


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,53 +1,101 @@
...
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isFormCorrectly = title.trim().length > 0
+    && imgUrl.trim().length > 0
+    && imdbUrl.trim().length > 0
+    && imdbId.trim().length > 0;
```
marina-tilniak, line 20: there should be check if both urls are valid, not only present

marina-tilniak: because now you can create movie with invalid urls

marina-tilniak (changes requested):

src/components/TextField/TextField.tsx
```diff
@@ -38,19 +43,23 @@ export const TextField: React.FC<Props> = ({
...
-          onBlur={() => setToched(true)}
+          onBlur={() => setTouched(true)}
         />
       </div>
 
       {hasError && (
         <p className="help is-danger">{`${label} is required`}</p>
       )}
+
+      {validUrl && (
```
TarasHoliuk, line 60: The name of this variable is not clear
if it is a boolean, let's name it `isUrlValid` and it should be true if the URL is valid

then render the error like here:
```
{!isUrlValid && (
  // error text
)} 
```

TarasHoliuk (changes requested):
Good job! Please check comments

marina-tilniak (approved):

PR #930, opened 2022-09-21
solution 1

kostya-kosyuk:
[DEMO LINK](https://kostya-kosyuk.github.io/react_movies-list-add-form/)


andrii-dmytruk (changes requested):
Fix enabling button when all required fields are filled

andrii-dmytruk (changes requested):
After adding 1 element, button is still active and if push it, the same element adds one more time.

PR #933, opened 2022-09-22
MmoviesListAddForm solution

NataliyaAn:
[DEMO LINK](https://NataliyaAn.github.io/react_tabs/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,99 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie:React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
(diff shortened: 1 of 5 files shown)


anna-kolmakova (changes requested):
For some reason in demo link is React-tabs task, please check

src/components/TextField/TextField.tsx
```diff
@@ -43,7 +43,7 @@ export const TextField: React.FC<Props> = ({
           type="text"
           placeholder={`Enter ${label}`}
           value={value}
-          onChange={event => onChange(event.target.value)}
+          onChange={event => onChange(event.target.value.trim())}
```
anna-kolmakova, line 46: You trim all values during input - that's why i can't add spaces to any field. I believe it shouldn't working like this. If you want to disable adding movie with empty data (spaces only), you can add trim to handleSubmit checker: title.trim() || discription.trim() || etc.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,99 @@
...
       <TextField
         name="imdbId"
         label="Imdb ID"
-        value=""
+        value={imdbId}
+        onChange={handleImdbId}
+        required
       />
 
       <div className="field is-grouped">
```
anna-kolmakova, line 99: Point 3 -  Disable the submit button until all the required fields are filled;

anna-kolmakova (changes requested):

NataliyaAn:
> > Чомусь у демо-посиланні є завдання React-tabs, будь ласка, перевірте
> 
> Я додав правильне посилання

PR #944, opened 2022-09-26
Solution of Movies list (Add form)

OleksiiPustovit:
[DEMO LINK](https://oleksiipustovit.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  const addMovie = (movie: Movie) => {
+    setMovies([...movies, movie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addMovie} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,86 @@
-import { useState } from 'react';
+import { useState, FormEvent } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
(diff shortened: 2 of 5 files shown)


src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  const addMovie = (movie: Movie) => {
+    setMovies([...movies, movie]);
+  };
```
VolodymyrChuyko, line 13:
```suggestion
  const handleAddMovie = (movie: Movie) => {
    setMovies(prevMovies => [...prevMovies, movie]);
  };
```
Follow the naming convention. Use an arrow function form of setMovies when operating with the previous value

src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addMovie} />
```
VolodymyrChuyko, line 21:
```suggestion
        <NewMovie onAdd={handleAddMovie} />
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,86 @@
...
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [count, setCount] = useState(0);
+
+  const handleSubmit = (event: FormEvent) => {
+    event.preventDefault();
+
+    setCount(count + 1);
```
VolodymyrChuyko, line 20:
```suggestion
    setCount(count => count + 1);
```

VolodymyrChuyko (changes requested):
Great job! Just small issues

src/components/NewMovie/NewMovie.tsx
OleksiiPustovit: Thanks for the feedback, I've made changes))

VolodymyrChuyko (approved):
Well done!

src/App.tsx
dellachaise: Agreed 👍

dellachaise (changes requested):
Great job 👍 

Only interested about usage `key` for form.
I asked because I use it in list kind of [here](https://reactjs.org/docs/lists-and-keys.html) but never in form. Where you heard, seen that you could use it this way and for what purpose?
Would be appreciate for information)

dellachaise (approved):
Well done 👍 

I saw the template(key for form was here), it was changed recently. And as I understand, we use it for tests 👌

OleksiiPustovit:
Very good question, I also thought about this when I saw the key in the form in the original task form.
After a long search, I still did not find the answer, after which I re-read the task and saw what it was for))

1. Errors should not be shown after clearing the form (change its key to
reinitialize the form);
 😃

PR #989, opened 2022-10-09
react: movies list add form

Alyona-Pisotska:
[DEMO LINK](https://Alyona-Pisotska.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,86 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie:React.FC<Props> = ({ onAdd }) => {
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,86 @@
...
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    onAdd({
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    });
```
Anastasiia-Svintsova, line 38: it will be better to create some const and use this const in the function.
```suggestion
   const movie = {
      title,
      description,
      imgUrl,
      imdbUrl,
      imdbId,
    }
    onAdd(movie);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,86 @@
...
+
+    setCount(state => state + 1);
+    resetField();
+  };
 
   return (
-    <form className="NewMovie" key={count}>
+    <form className="NewMovie" key={count} onSubmit={handleSubmit}>
```
Anastasiia-Svintsova, line 45: Why do you need key here? Delete it. And you also don't need counter

Anastasiia-Svintsova (changes requested):
Good job! Just fix some comments

Anastasiia-Svintsova (approved):

PR #993, opened 2022-10-12
solution

ihorandrianov:
[DEMO LINK](https://ihorandrianov.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,7 +24,7 @@ export const App = () => {
         <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie onAdd={handleAdd} />
+        <NewMovie onAdd={handleAdd} movies={movies} />
       </div>
     </div>
   );
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,30 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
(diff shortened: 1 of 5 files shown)


naziamoff (changes requested):
- You should clear the form fields after the submit
- I can add one movie a lot of times, you should validate it
- I can put two spaces as a movie name. You should handle this too

src/App.tsx
```diff
@@ -1,16 +1,30 @@
...
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  const handleAdd = (movie: Movie) => {
+    setMovies((prevState) => {
+      const newState = [...prevState];
+
+      newState.push(movie);
```
Denchepornyuk, line 15: don't use push, add movie with array copy

src/components/Button.tsx
```diff
@@ -0,0 +1,22 @@
+import { FC } from 'react';
+import { useFormikContext } from 'formik';
+import { Movie } from '../types/Movie';
+
+export const Button: FC = () => {
+  const ctx = useFormikContext<Movie>();
```
Denchepornyuk, line 6: 🔥

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,58 +1,92 @@
...
+          .required(),
+        imdbId: Yup.string()
+          .required()
+          .whitespace('Write some id here!'),
+      })}
+      onSubmit={(values, { resetForm }) => {
+        onAdd(values);
+        resetForm();
+      }}
+    >
```
Denchepornyuk, line 59: I didn't work with formik, we use React Final Form in Mate but it looks great 🚀

Denchepornyuk (approved):

naziamoff (approved):

PR #994, opened 2022-10-15
react movies-list-add-form is done

MaxShylo:
[DEMO LINK](https://MaxShylo.github.io/react_movies-list-add-form/)


Anastasiia-Svintsova (changes requested):
Think about how to rewrite your validator. Because now I can add movies without the required fields
Animation

qoa11a (approved):

PR #995, opened 2022-10-18
movie list form implented

illia-sosnytskyi:
[DEMO LINK](https://illia-sosnytskyi.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,94 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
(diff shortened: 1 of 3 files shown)


AntonVaida (approved):

sergii-nosachenko (changes requested):
You need to check if required fields values are not consist of spaces. Use trim method.
Screenshot 2022-10-31 at 17 01 27

src/components/NewMovie/NewMovie.tsx
```diff
@@ -25,19 +23,21 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
       imdbId,
     });
 
-    setCount(num => num + 1);
-
     setTitle('');
     setDescrption('');
     setImgUrl('');
     setImdbUrl('');
     setImdbId('');
```
sergii-nosachenko, line 30:
```suggestion
```
Unnecessary code, when you change key of this component it will be rendered with default values.

sergii-nosachenko (changes requested):
Small fix and it will be perfect.

sergii-nosachenko (approved):
Perfect 🔥

PR #1001, opened 2022-10-21
movies list has been implemented

vmamchur:
[DEMO LINK](https://vmamchur.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,10 +1,35 @@
 import { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (newMovie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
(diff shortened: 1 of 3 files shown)


imaksweb (approved):
Nice work 👾

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(event) => {
+          setTitle(event);
```
Anna-Tkachenko, line 51: set event.target.value to state variables, not event

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +98,12 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={
+              !title.trim()
+                || !imgUrl.trim()
+                || !imdbUrl.trim()
+                || !imdbId.trim()
```
Anna-Tkachenko, line 105: create separate variable for this check and use this variable for disabled attribute

Anna-Tkachenko (changes requested):

src/components/NewMovie/NewMovie.tsx
vmamchur: the callback onChange seems like it's already a string. so i'll use it directly.
image

Anna-Tkachenko (approved):
⭐️

PR #1002, opened 2022-10-21
solution was added

Maksym845:
- [DEMO LINK](https://Maksym845.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,12 +1,12 @@
-import { useState } from 'react';
+import { FC, FormEvent, useState } from 'react';
 import { TextField } from '../TextField';
 import { Movie } from '../../types/Movie';
 
-interface OnAdd {
+interface Props {
   onAdd: (movie: Movie) => void;
 }
 
-export const NewMovie = ({ onAdd }: OnAdd) => {
+export const NewMovie: FC<Props> = ({ onAdd }) => {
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +66,21 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={((title === ''
+              || imgUrl === ''
+              || imdbUrl === ''
+              || imdbId === '') && true) || false}
```
Yeliseiev, line 72: Better create const `isRequired` for this check, also according the checklist `make sure you can't add movie with empty data (with spaces only);`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +66,21 @@ export const NewMovie = () => {
...
+            onClick={(event) => {
+              event.preventDefault();
+              onAdd({
+                title,
+                description,
+                imgUrl,
+                imdbUrl,
+                imdbId,
+              });
+              setCount((prevCount) => prevCount + 1);
```
Yeliseiev, line 82: Try to keep the event handler as a separate function.
And you must clear the form after submit.

Yeliseiev (changes requested):
Great job! But you need to make some changes according to the checklist and readme file.

PR #1012, opened 2022-10-24
solution

myroslav-diiak:
[DEMO LINK](https://myroslav-diiak.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,26 @@
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
-export const App = () => {
+export const App: React.FC = () => {
+  const [moviesList, setMoviesList] = useState(moviesFromServer);
+
+  const onAdd = (movie: Movie) => {
+    setMoviesList(currentMovies => [...currentMovies, movie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={moviesList} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie
+          onAdd={onAdd}
+        />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,98 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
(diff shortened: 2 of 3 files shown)


src/App.tsx
```diff
@@ -1,16 +1,26 @@
...
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+
+export const App: React.FC = () => {
+  const [moviesList, setMoviesList] = useState(moviesFromServer);
+
+  const onAdd = (movie: Movie) => {
+    setMoviesList(currentMovies => [...currentMovies, movie]);
```
vladimir-dublya, line 12: I am not sure, but i am think that will be better if you will use callback here

vladimir-dublya (approved):

src/App.tsx
Denchepornyuk: useCallback would be good here, but we haven't learned it yet 🙂

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,98 @@
...
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
```
Denchepornyuk, line 12: you don't need count for this task

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,98 @@
...
+      onSubmit={(event) => {
+        event.preventDefault();
+
+        setCount(count + 1);
+
+        onAdd({
+          title,
+          description,
+          imgUrl,
+          imdbUrl,
```
Denchepornyuk, line 36: it is better to move the logic from the submit button to a separate method, a separate method can be optimized using the Callback hook

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,98 @@
...
+          imdbId,
+        });
+
+        setTitle('');
+        setDescription('');
+        setImgUrl('');
+        setImdbUrl('');
+        setImdbId('');
+      }}
```
Denchepornyuk, line 45: Create a separate method for cleaning the form

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +101,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={inputsValidation()}
```
Denchepornyuk, line 104: nope, you should use onBlur for checking inputs 🙂

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1014, opened 2022-10-24
implemented form movies-list

nikitaovch99:
[DEMO LINK](https://nikitaovch99.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,7 +9,7 @@ export const App = () => {
   const [totalFilm, setTotalFilm] = useState(moviesFromServer);
   const addFilm = (newFilm: Movie) => {
     setTotalFilm(
-      [...moviesFromServer, newFilm],
+      [...totalFilm, newFilm],
     );
   };
 


src/App.tsx
```diff
@@ -1,16 +1,25 @@
...
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [totalFilm, setTotalFilm] = useState(moviesFromServer);
+  const addFilm = (newFilm: Movie) => {
+    setTotalFilm(
+      [...moviesFromServer, newFilm],
+    );
```
igor-1987, line 13: I guess you have to add new movie to the items in the state, rather then to initial list from file.

src/App.tsx
```diff
@@ -1,16 +1,25 @@
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [totalFilm, setTotalFilm] = useState(moviesFromServer);
```
igor-1987, line 9: It makes more sense to call it in plural form or to use something like "filmList", "movieList"

igor-1987 (commented):
GJ! There are couple left tiny things to fix.

src/App.tsx
naziamoff: I agree with @igor-1987 that this naming does not seem to be clear at all
Most of all I like the option `moviesList`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,44 @@
...
+
+type Event = {
+  target: { name: string,
+    value: string,
+  }
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
```
naziamoff, line 17: don't leave unnecessary comments

naziamoff (approved):
GJ, but I still recommend you to implement the validation

PR #1019, opened 2022-10-25
movie list without url pattern check

TheVovchik:
**[DEMO LINK](https:/TheVovchik.github.io/react_movies-list-add-form/)**


src/App.tsx
```diff
@@ -1,16 +1,26 @@
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
```
Anna-Borodaienko, line 9: moviesFromServer is not primitive, add type for useState

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,101 @@
...
+    setImdbUrl('');
+    setImdbId('');
+
+    setCount(currentCount => (
+      currentCount + 1
+    ));
+  };
+
+  const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
```
Anna-Borodaienko, line 32: What is ```e```? Maybe it would be better to name it ```event```

Anna-Borodaienko (changes requested):

src/components/NewMovie/NewMovie.tsx
TheVovchik: e: React.FormEvent<HTMLFormElement>
what is cn in classNames =) it's a shorthand naming)

Anna-Borodaienko (approved):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,101 @@
...
+  const resetInputs = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+
+    setCount(currentCount => (
+      currentCount + 1
+    ));
```
Anna-Tkachenko, line 28: this functionality shouldn't be inside resetInputs function, it doesn't reset any input

Anna-Tkachenko (approved):

PR #1033, opened 2022-10-26
Implement movies list with adding new movie

imaksweb:
- [DEMO LINK](https://imaksweb.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,30 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  const addMovie = (movie: Movie) => {
+    setMovies(currentMovies => {
+      const isExist = currentMovies.find(
+        currentMovie => currentMovie.imdbId === movie.imdbId,
+      );
+
+      return isExist ? currentMovies : [...currentMovies, movie];
+    });
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,96 @@
...
+
+        onAdd({
+          title,
+          description,
+          imgUrl,
+          imdbUrl,
+          imdbId,
+        });
+        clearForm();
+        setCount(current => current + 1);
```
Vladosik1337, line 42: It might be better to put it in a function

Vladosik1337 (approved):
Nice

src/App.tsx
```diff
@@ -1,16 +1,30 @@
...
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  const addMovie = (movie: Movie) => {
+    setMovies(currentMovies => {
+      const isExist = currentMovies.find(
```
Anna-Tkachenko, line 13: isExist - name for some boolean variable, method find doesn't return boolean value
think about variable name or change method

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,103 @@
...
+        onAdd({
+          title: title.trim(),
+          description: description.trim(),
+          imgUrl: imgUrl.trim(),
+          imdbUrl: imdbUrl.trim(),
+          imdbId: imdbId.trim(),
+        });
+        clearForm();
+        increaseCount();
+      }}
```
Anna-Tkachenko, line 50: create separate function for submit handler

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,103 @@
...
+        }}
         required
       />
 
       <TextField
         name="description"
         label="Description"
-        value=""
+        value={description}
+        onChange={(value) => {
```
Anna-Tkachenko, line 68: it's not value, it's event, check variable name

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #1038, opened 2022-10-26
Implement movies list with form

anton-iskryk:
[DEMO LINK](https://anton-iskryk.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,95 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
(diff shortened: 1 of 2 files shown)


AniaHna (commented):
nice!

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +98,12 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={
+              !title.trim()
+                || !imgUrl.trim()
+                || !imdbUrl.trim()
+                || !imdbId.trim()
+            }
           >
```
naziamoff, line 107: looks like this should be a variable`shouldDisableSubmitButton` or sth like that

naziamoff (changes requested):
there's no link validation 😢

naziamoff (approved):
Found out that validation is optional
Nevertheless, I recommend you to implement it

PR #1046, opened 2022-10-27
task solution

vladyslav-harkusha:
[DEMO LINK](https://vladyslav-harkusha.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,9 +8,12 @@ import { Movie } from './types/Movie';
 
 export const App: FC = () => {
   const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+  const [count, setCount] = useState(0);
 
   const addMovie = (movie: Movie) => {
     setMovies(currentMovies => [...currentMovies, movie]);
+
+    setCount(current => current + 1);
   };
 
   return (
@@ -19,7 +22,7 @@ export const App: FC = () => {
         <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie onAdd={addMovie} />
+        <NewMovie onAdd={addMovie} key={count} />
       </div>
     </div>
   );
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,25 @@
+import { FC, useState } from 'react';
+
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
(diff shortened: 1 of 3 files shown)


Yuriy-Komolov (approved):

Yuriy-Komolov:
Good Job!

src/App.tsx
```diff
@@ -1,16 +1,25 @@
...
-export const App = () => {
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addMovie} />
```
sergii-nosachenko, line 22:
```suggestion
        <NewMovie onAdd={addMovie} key={count} />
```
Move counter logic here. When you'll increase count your component will be created from scratch with default values and no errors.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,87 @@
...
+      imdbId,
+    });
+
+    setTitle('');
+    setDesription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+
+    setCount(current => current + 1);
```
sergii-nosachenko, line 35:
```suggestion
```
Move setCount to parent component and use it in onAdd handler. You don't need to reset your form.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,87 @@
...
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={value => setTitle(value)}
```
sergii-nosachenko, line 52:
```suggestion
        onChange={setTitle}
```
There is no event here. TextField component calls your handler with field value, so you don't need wrapper here.

Fix it everywhere.

sergii-nosachenko (changes requested):
You need to add additional check if your fields consists of spaces only.
Screenshot 2022-10-31 at 17 15 53

sergii-nosachenko (approved):
Perfect 🔥

PR #1047, opened 2022-10-27
Movies list add form

Yuriy-Komolov:
[DEMO LINK](https://Yuriy-Komolov.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,90 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (movie: Movie) => void
+}
+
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,90 @@
...
+      imdbId,
+    });
+
+    setCount(prevCount => prevCount + 1);
+
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
```
Anna-Tkachenko, line 32: will be better to create separate function that will clear the form

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,90 @@
...
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={setTitle}
```
Anna-Tkachenko, line 55: you should handle event in onChange handler and set target value to the state, for now I don't see this functionality

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #1064, opened 2022-11-08
solution

AnatoliiMazepa:
[DEMO LINK](https://anatoliimazepa.github.io/react_movies-list-add-form/)


liliia-livandovskaya (changes requested):
Add validation to these fields so the user could not insert nonlink values Screenshot 2022-11-09 at 17 41 52

vlad-demchuk (approved):

PR #1067, opened 2022-11-09
solution

mykola8426:
https://mykola8426.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,105 @@
 import { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,105 @@
...
       <TextField
         name="imgUrl"
         label="Image URL"
-        value=""
+        value={newMovie.imgUrl}
+        onChange={(value) => {
+          setNewMovie({
+            ...newMovie,
+            imgUrl: value,
+          });
```
ArtemTopolskyi, line 74: looks like this logic is repeated several times, lets move it to separate function

ArtemTopolskyi (approved):

PR #1072, opened 2022-11-15
solution

YuriMoreau:
[DEMO LINK](https://yurimoreau.github.io/react_movies-list-add-form/)


Anastasiia-Svintsova (changes requested):
- I shouldn't be able to add empty fields 
Animation

PR #1094, opened 2022-11-30
implemented add movie to list

Artem66:
[DEMO LINK](https://Artem66.github.io/react_movies-list-add-form/)


Illia-Skladnik (approved):
Good work! 
A few ideas about code.
1.  'Title' Shouldn't work if you pass by '  ' spaces only. Maybe you should trim method for that.
2.  In URLs fields should be required validation of correct URL.

mykola-koval (approved):

PR #1095, opened 2022-11-30
Implement movies add form

anastasiia-tilikina:
[DEMO LINK](https://anastasiia-tilikina.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,22 @@
 import './App.scss';
-import { MoviesList } from './components/MoviesList';
+import { useState } from 'react';
 import { NewMovie } from './components/NewMovie';
+import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App = () => {
+  const [moviesForRender, setMoviesForRender] = useState(moviesFromServer);
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={moviesForRender} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie) => {
+          setMoviesForRender(curr => [...curr, movie]);
+        }}
+        />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -14,9 +14,16 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   const [imdbUrl, setImdbUrl] = useState('');
   const [imdbId, setImdbId] = useState('');
 
+  const isValidUrl = (url: string) => {
+    // eslint-disable-next-line max-len
+    const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@.,\w_]*)#?(?:[.!/\\\w]*))?)$/;
+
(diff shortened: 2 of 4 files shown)


src/App.tsx
```diff
@@ -1,16 +1,22 @@
...
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={moviesForRender} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie) => {
+          setMoviesForRender(curr => [...curr, movie]);
```
vlad-demchuk, line 17: let's move this function to a separate variable

src/components/TextField/TextField.tsx
```diff
@@ -18,14 +19,15 @@ export const TextField: React.FC<Props> = ({
...
 }) => {
-  // generage a unique id once on component load
   const [id] = useState(() => `${name}-${getRandomDigits()}`);
 
-  // To show errors only if the field was touched (onBlur)
   const [touched, setToched] = useState(false);
-  const hasError = touched && required && !value;
+  const hasError = touched && required && !value.trim();
+
+  const isNotValid = !isValidUrl && !hasError && touched;
```
vlad-demchuk, line 30: use state for errors

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,98 @@
...
+  const [imdbId, setImdbId] = useState('');
+
+  const isValidUrl = (url: string) => {
+    // eslint-disable-next-line max-len
+    const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@.,\w_]*)#?(?:[.!/\\\w]*))?)$/;
+
+    return pattern.test(url);
+  };
+
+  const isSubmitDisabled = !title.trim()
```
vlad-demchuk, line 24: use state for errors and check it in `handleSubmit`

vlad-demchuk (changes requested):
you don't hide an error for this field
image

anastasiia-tilikina:
I don't understand why I need to use state for error checks.
Since I don't need to rerender components when an error occurs.
When the text field changes my component rerenders and I have another value in var (title, URL,  etc.)
So I only need to create var, like 'isSubmitDisabled' to check those values, which are different on every rerender.
If I use state instead, I will need to write at least 4 additional functions to change them.
And I need 'isSubmitDisabled' anyway to disable the button. The only difference is that I will use in it variables from state.
And I don't need to check errors on submit since my button is disabled when an error occurs and submit is not possible.
So I don't need an extra check on submission.

Or maybe You mean something different in those comments, which I didn't grasp?)

AlisaVlasova (changes requested):
image
error on this field still doesn't disappear( 
did you deploy your changes

anastasiia-tilikina:
Sorry, my bad)
Seems like I really forgot to deploy)

Now It's done!
I've checked it twice )

vlad-demchuk:
> I don't understand why I need to use state for error checks. Since I don't need to rerender components when an error occurs. When the text field changes my component rerenders and I have another value in var (title, URL, etc.) So I only need to create var, like 'isSubmitDisabled' to check those values, which are different on every rerender. If I use state instead, I will need to write at least 4 additional functions to change them. And I need 'isSubmitDisabled' anyway to disable the button. The only difference is that I will use in it variables from state. And I don't need to check errors on submit since my button is disabled when an error occurs and submit is not possible. So I don't need an extra check on submission.
> 
> Or maybe You mean something different in those comments, which I didn't grasp?)

it's true. If you don't need to change you can avoid using state. just thought you changed it, but if it works without state - good

vlad-demchuk (approved):

PR #1112, opened 2022-12-05
the react_movies-list-add-form is done

Anastasiya145:
[DEMO LINK](https://Anastasiya145.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,96 @@
 import { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isDisabled
+  = !title.trim()
+  || !imgUrl.trim()
+  || !imdbId.trim()
+  || !imdbUrl.trim();
+
+  const clearForm = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
+
+  const FormSubmit = (ev: React.FormEvent<HTMLFormElement>) => {
+    ev.preventDefault();
+
+    const movieToAdd = {
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,96 @@
...
+
+  const clearForm = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
+
+  const FormSubmit = (ev: React.FormEvent<HTMLFormElement>) => {
```
tetiana-ivanchenko, line 31:
```suggestion
  const formSubmit = (event: React.FormEvent<HTMLFormElement>) => {
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,96 @@
...
+        value={description}
+        onChange={(newDescription) => setDescription(newDescription)}
       />
 
       <TextField
         name="imgUrl"
         label="Image URL"
-        value=""
+        value={imgUrl}
+        onChange={(newImgUrl) => setImgUrl(newImgUrl)}
```
tetiana-ivanchenko, line 76:
```suggestion
        onChange={setImgUrl}
```

tetiana-ivanchenko: fix it everywhere

tetiana-ivanchenko (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,96 @@
...
+  || !imdbId.trim()
+  || !imdbUrl.trim();
+
+  const clearForm = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
```
OlEzhA0, line 29: you can wrap it with useCallback

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,96 @@
...
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
OlEzhA0, line 15: it's ok to use a lot of states but you can create only, keep there movie object and change fields inside

OlEzhA0 (commented):
Hey! Very good job
I left few comments so you can check how do you can manage your state with 1 `useState`
Your solution is also good but can be improved. If you decide to leave it like here, that's ok

OlEzhA0 (changes requested):
check comments above :)

src/components/NewMovie/NewMovie.tsx
AlisaVlasova: on QnA we decided that it`ll be better to keep separate states in this case, and not change TextField logic that returns onChange string value instead of event

AlisaVlasova (approved):

PR #1115, opened 2022-12-05
Make movie list add form v1.0

bohdan-stepanets:
[DEMO LINK](https://bohdan-stepanets.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -6,17 +6,7 @@ type Props = {
   onAdd: (movie: Movie) => void,
 };
 
-// export interface Movie {
-//   title: string;
-//   description: string;
-//   imgUrl: string;
-//   imdbUrl: string;
-//   imdbId: string;
-// }
-
 export const NewMovie: React.FC<Props> = ({ onAdd }) => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
   const [count, setCount] = useState(0);
   const [title, setTitle] = useState('');
   const [description, setDescription] = useState('');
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,88 @@
...
+  const onFormSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    onAdd(newMovie);
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+    setCount(prevCount => prevCount + 1);
+  };
```
kryvets-vadym, line 34: here you should separate everything with empty lines depending on the context for better readability, you should make a separate function for resetting the form and use it here

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,88 @@
...
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
```
kryvets-vadym, line 7: perhaps using an interface to typify props is more appropriate here

kryvets-vadym (approved):
Good work, but minor edits need to be made)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+    event.preventDefault();
+
+    onAdd(newMovie);
+
+    resetForm();
+
+    setCount(prevCount => prevCount + 1);
+  };
+
+  const buttonActive = title.trim().length > 0
```
denys-danyliuk, line 43:
```suggestion
  const isButtonActive = title.trim()
```
it will work this way

denys-danyliuk (changes requested):
add an error if field filled with spaces
Знімок екрана 2022-12-07 о 10 45 59
Знімок екрана 2022-12-07 о 10 53 32

PR #1133, opened 2022-12-09
all done

Yevhen-Pavlovskyi:
https://yevhen-pavlovskyi.github.io/react_movies-list-add-form/


Anastasiia-Svintsova (changes requested):
- What do I need to do to be able to add the movie? Fix this or add some notification about why the user can't add anything if all fields are filled.

https://user-images.githubusercontent.com/99403198/206855313-9863c400-6102-4232-92d0-4a9c7f3012cf.mov

qoa11a (approved):

PR #1134, opened 2022-12-13
Movies list add form task solution

juliiaap:
- [DEMO LINK](https://juliiaap.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,121 @@
-import { useState } from 'react';
+import './NewMovie.scss';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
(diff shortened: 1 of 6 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,121 @@
...
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const defaultMovie = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
```
Bohdan-Maksymiuk, line 23: such variable better to store outside the components
cause on every rerender there will be new object an new link on it

Bohdan-Maksymiuk (approved):

src/components/NewMovie/NewMovie.tsx
juliiaap: thank you! fixed it

PR #1137, opened 2022-12-17
solution

SergIats:
[DEMO LINK](https://SergIats.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,91 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,91 @@
...
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(value) => setTitle(value)}
```
tetiana-ivanchenko, line 56:
```suggestion
        onChange={setTitle}
```

tetiana-ivanchenko: fix everywhere

tetiana-ivanchenko (changes requested):

src/components/NewMovie/NewMovie.tsx
SergIats: Tetiana,hi! Thank you for the review. You're right - without using previous value, your variant is better. I've done it.

nazarmatsevych (approved):

PR #1157, opened 2023-01-05
solution

lefteris-L:
[DEMO LINK](https://lefteris-l.github.io/react_movies-list-add-form/)


meforoma (changes requested):
you're not really validating the url - even a single char would pass the validation
check it and make sure to properly validate

naziamoff (approved):
gj!
Would be great if it did not allow to add the same movie twice

PR #1163, opened 2023-01-09
react_movies-list-add-form

kolya-movchan:
[DEMO LINK](https://kolya-movchan.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -61,7 +61,7 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
     <form
       className="NewMovie"
       key={count}
-      onSubmit={(event) => handleSubmit(event)}
+      onSubmit={handleSubmit}
     >
       <h2 className="title">Add a movie</h2>
 


tetiana-ivanchenko (changes requested):
GJ, just one amendment:
if you want to add link validation - it's not the best approach  to clear this field and add a message that it's required. better to show some clarifying message, like "Please, enter a valid URI" or smth like that

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,107 @@
...
+    onAdd(newMovie);
+    reset();
+  };
 
   return (
-    <form className="NewMovie" key={count}>
+    <form
+      className="NewMovie"
+      key={count}
+      onSubmit={(event) => handleSubmit(event)}
```
tetiana-ivanchenko, line 64:
```suggestion
      onSubmit={handleSubmit}
```

tetiana-ivanchenko (changes requested):

nazarmatsevych (approved):

PR #1170, opened 2023-01-13
solution

Fynkzia:
[DEMO LINK](https://Fynkzia.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useCallback, memo } from 'react';
 import { TextField } from '../TextField';
 import { Movie } from '../../types/Movie';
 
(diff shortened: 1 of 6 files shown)


TetKopot (changes requested):
gj! but the user should be able to add just a valid URL

Fynkzia:
> gj! but the user should be able to add just a valid URL

This is an optional task

TetKopot (approved):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,115 @@
...
+    setImgUrlValue(value);
+  };
+
+  const handleImdbUrlChange = (value: string) => {
+    setImdbUrlValue(value);
+  };
+
+  const handleImdbIdChange = (value: string) => {
+    setImdbIdValue(value);
+  };
```
solaryasha, line 37: Let's use `useCallback` for render optimisation, because you are passing these callbacks to another components

solaryasha (changes requested):

solaryasha (approved):
👍

PR #1174, opened 2023-01-14
solution

batstolya:
[DEMO LINK](https://batstolya.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,87 @@
...
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    };
+
+    if (!isValidForm) {
+      onAdd(newMovie);
+    }
```
Oleksandr-Filo, line 34: Strange logic - if form is not valid we add a new movie. Probably you should change variables names or logic inside validation or something else.

Oleksandr-Filo (approved):
As for me it is better to add the ability to validate form for spaces only - make sure you can't add movie with empty data (with spaces only). If you want you can add this.
image

src/components/NewMovie/NewMovie.tsx
batstolya: good point! thanks

ivan-kadykalo (changes requested):
I can add empty film
image

Imdb url doesn't clearing after the submit

ivan-kadykalo (approved):
Good job 🔥

PR #1186, opened 2023-01-15
solution of react movies list add form

RedkoSerhii:
[DEMO LINK](https://RedkoSerhii.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,27 @@
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+import { Movie } from './types/Movie';
 import moviesFromServer from './api/movies.json';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addNewFilm = (movie: Movie) => {
+    setMovies(current => [
+      ...current,
+      movie,
+    ]);
+  };
+
   return (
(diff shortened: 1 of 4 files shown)


NadiiaKoch (approved):
GJ!

src/App.tsx
```diff
@@ -1,16 +1,27 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+import { Movie } from './types/Movie';
 import moviesFromServer from './api/movies.json';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addNewFilm = (movie: Movie) => {
```
solaryasha, line 11: You should wrap this callback with `useCallback` , cause you are passing the function to another component

src/App.tsx
```diff
@@ -1,16 +1,27 @@
...
+    setMovies(current => [
```
solaryasha, line 12:
```suggestion
    setMovies(currentMovies => [
```

solaryasha (changes requested):

src/App.tsx
RedkoSerhii: fixed it

src/App.tsx
RedkoSerhii: fixed it

solaryasha (approved):
Nice!

PR #1195, opened 2023-01-15
React movies list add form

oleh-yablunovskyi:
[DEMO LINK](https://oleh-yablunovskyi.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,27 @@
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
-export const App = () => {
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = (newMovie: Movie) => {
+    setMovies((prevMovies => (
+      [...prevMovies, newMovie]
+    )));
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
+
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addMovie} />
       </div>
     </div>
   );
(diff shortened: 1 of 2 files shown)


romanblyashuk94 (approved):
Good job 👍

src/App.tsx
```diff
@@ -1,16 +1,27 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = (newMovie: Movie) => {
```
solaryasha, line 11: You should wrap this callback with `useCallback` , cause you are passing the function to another component

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,110 @@
...
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (newMovie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
```
solaryasha, line 10: What about the `React.memo` ?

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,110 @@
...
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={newTitle}
         required
+        onChange={(value: string) => setNewTitle(value)}
```
solaryasha, line 74: You have two options:

2. create a specific function that takes value and triggers `setState` function with that value. But you need to wrap this function with useCallback (and you will get render optimization)

solaryasha (changes requested):

src/App.tsx
oleh-yablunovskyi: Done

src/components/NewMovie/NewMovie.tsx
oleh-yablunovskyi: Done

src/components/NewMovie/NewMovie.tsx
oleh-yablunovskyi: Done. Thanks a lot for the detailed explanation!

solaryasha (approved):
Good job!

PR #1196, opened 2023-01-15
Solution

4ebupel:
[DEMO LINK](https://4ebupel.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movie, setMovie] = useState(moviesFromServer);
+
+  const onAdd = (newMovie: Movie) => {
+    setMovie(prevMovie => [...prevMovie, newMovie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movie} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={onAdd} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,101 @@
-import { useState } from 'react';
+import { FC, useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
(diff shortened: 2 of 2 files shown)


AntonMartynes (approved):
good job

src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movie, setMovie] = useState(moviesFromServer);
+
+  const onAdd = (newMovie: Movie) => {
```
solaryasha, line 11: Let's use `useCallback` for render optimisation

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,101 @@
...
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(event) => {
+          setTitle(event);
+        }}
```
solaryasha, line 58: Don't use inliners, with `useCallback` you could accomplish some rendering optimization

solaryasha (changes requested):

src/App.tsx
4ebupel: Done

src/components/NewMovie/NewMovie.tsx
4ebupel: I don't understand what exactly you want me to do, do you want me to implement a "debouncer" or what? Because as far as I understand, we use `useCallback` only in two cases: when we "pass" some func to the child component or when we making a debounce decorator

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,107 @@
...
+};
+
+// const debounce = (func: () => void, delay: number) => {
+//   let id: number;
+
+//   return (...args: unknown[]) => {
+//     clearTimeout(id);
+//     id = setTimeout(func, delay, ...args);
+//   };
+// };
```
solaryasha, line 16: Don't leave options

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,107 @@
...
+  const [imgUrl, setImgUrl] = useState(defaultValue);
+  const [imdbUrl, setImdbUrl] = useState(defaultValue);
+  const [imdbId, setImdbId] = useState(defaultValue);
+
+  const changeTitle = useCallback((value: string) => setTitle(value), []);
+  // eslint-disable-next-line max-len
+  const changeDescription = useCallback((value: string) => setDescription(value), []);
+  const changeImgUrl = useCallback((value: string) => setImgUrl(value), []);
+  const changeImdbUrl = useCallback((value: string) => setImdbUrl(value), []);
+  const changeImdbId = useCallback((value: string) => setImdbId(value), []);
```
solaryasha, line 32: You have two options:

2. create a specific function that takes value and triggers `setState` function with that value. But you need to wrap this function with useCallback (and you will get render optimization) - that what you actually done and it's okay

solaryasha (approved):

PR #1198, opened 2023-01-15
solution

DiDemkiv:
[DEMO LINK](https://DiDemkiv.github.io/react_movies-list-add-form/)


batstolya:
after adding movie , fields should not show error


image

batstolya:
you need to think over the logic so that the user cannot add a movie by filling in the fields with empty strings
image

PR #1200, opened 2023-01-15
solution

victoria-kovalenko:


kreisberg (approved):
Seems great!

Have same mistakes with test. But it's actually working, was confirmed on tech-check.

ivan-kadykalo (changes requested):
> NewMovie should check if title, imgUrl, imdbUrl, imdbId are entered when an input looses focus (onBlur) and show an error and a red border if needed
It should work for all fields (except description) linked it is for the title

ivan-kadykalo (approved):
Good job 🔥

PR #1203, opened 2023-01-16
solution

AntonMartynes:
https://AntonMartynes.github.io/react_movies-list-add-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,25 @@
+import { FC, useState } from 'react';
+
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
-export const App = () => {
+export const App:FC = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = (newMovie: Movie) => {
+    setMovies(prev => [...prev, newMovie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addMovie} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,116 @@
-import { useState } from 'react';
+import { FC, FormEvent, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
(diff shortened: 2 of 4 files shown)


Anastasia4sik (approved):
For me everything looks nice

src/App.tsx
```diff
@@ -1,16 +1,25 @@
...
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+
+export const App:FC = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = (newMovie: Movie) => {
+    setMovies(prev => [...prev, newMovie]);
```
solaryasha, line 13:
```suggestion
    setMovies(prevMovies => [...prev, newMovie]);
```

src/App.tsx
```diff
@@ -1,16 +1,25 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+
+export const App:FC = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = (newMovie: Movie) => {
```
solaryasha, line 12: Let's use `useCallback` for rendering optimization, cause you will pass this callback to the another component

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,116 @@
...
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (newMovie: Movie) => void
+};
+
+export const NewMovie: FC<Props> = ({ onAdd }) => {
```
solaryasha, line 9: What about the `React.memo` ?

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,116 @@
...
+    setDescription('');
+    setImgUrl('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+
+    setCount(prev => prev + 1);
+  };
+
+  const isPossibleToSubmit = title && imgUrl && imdbUrl && imdbId;
```
solaryasha, line 53: What if I pass `          ` as a title value ? You should check for that case

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,116 @@
...
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={value => {
+          setTitle(value);
+        }}
```
solaryasha, line 69: You have two options:

2. create a specific function that takes value and triggers `setState` function with that value. But you need to wrap this function with useCallback (and you will get render optimization

solaryasha (changes requested):

valerii-derkach (approved):

PR #1213, opened 2023-01-18
solution

Lun4trick:
[DEMO LINK](https://lun4trick.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,93 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+type Props = {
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,93 @@
...
+        value={newDescription}
+        onChange={setNewDescription}
       />
 
       <TextField
         name="imgUrl"
         label="Image URL"
-        value=""
+        value={newImgUrl}
+        onChange={setNewImgUrl}
```
nurdanokcu, line 73:
```suggestion
        onChange={(newValue) => setNewImgUrl(newValue)}
```
I know that it works the other way also but this makes it more understandable.

nurdanokcu (approved):
Url validation would be a nice addition but it is optional so I am approving.

naziamoff (approved):
well done. Would be great if you added links validation

PR #1226, opened 2023-01-25
solution

lukacsja:
[DEMO LINK](https://lukacsja.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,93 @@
 import { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+interface Props {
+  onAdd: (movie: Movie) => void,
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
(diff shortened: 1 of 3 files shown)


lefteris-L (approved):
Awesome job 👍

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,93 @@
...
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+
+    setCount(count + 1);
+  };
+
+  const isFormValid = () => (
```
naziamoff, line 39:
```suggestion
  const isFormValid = 
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +96,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!isFormValid()}
```
naziamoff, line 99: is there a reason to have it as a function?
```suggestion
            disabled={!isFormValid}
```

naziamoff (approved):

src/components/NewMovie/NewMovie.tsx
lukacsja: actually no, good point. probably because i have been heavily involved in writing functions recently 😄

PR #1227, opened 2023-01-25
finished

jackelashkar:
[DEMO LINK](https://jackelashkar.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,94 @@
 import { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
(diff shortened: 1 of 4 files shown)


naziamoff (changes requested):
i can submit the form if I enter a space in each field

jackelashkar:
> i can submit the form if I enter a space in each field

Done!

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +97,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!isValid.trim()}
```
naziamoff, line 100: how do you expect this one to work?

naziamoff (changes requested):
submit is still available with empty spaces for required fields
Screenshot 2023-01-27 at 12 45 58

jackelashkar:
> submit is still available with empty spaces for required fields Screenshot 2023-01-27 at 12 45 58

I fixed it now :)

naziamoff (approved):
would be great to display the error why I can't submit the form. I mean that the title is required, instead of just disabling button. Not really user-friendly this way

jackelashkar:
> would be great to display the error why I can't submit the form. I mean that the title is required, instead of just disabling button. Not really user-friendly this way

yes this was optional thing in the task and I didn't know how to do it unforunately

PR #1229, opened 2023-01-26
solution

Iksich:
[DEMO LINK](https://Iksich.github.io/react_movies-list-add-form/)


naziamoff (changes requested):
That's from readme:

NewMovie should check if title, imgUrl, imdbUrl, imdbId are entered when an input looses focus (onBlur) **and show an error and a red border if needed** (learn how it it implemented in the TextField);

naziamoff (commented):
oh, disregard the last point, it works. Looking further

naziamoff (approved):
code is also good 👍

PR #1231, opened 2023-01-27
done

alaaashkar:
[DEMO LINK](https://alaaashkar.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,50 +1,109 @@
 import { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (newMovie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,109 @@
...
+  const [imgUrl, setimgUrl] = useState('');
+  const [imdbUrl, setimdbUrl] = useState('');
+  const [imdbId, setimdbId] = useState('');
+
+  const clear = () => {
+    setTitle('');
+    setDescription('');
+    setimgUrl('');
+    setimdbUrl('');
+  };
```
meforoma, line 24: missed one
```suggestion
  const clear = () => {
    setTitle('');
    setDescription('');
    setimgUrl('');
    setimdbUrl('');
    setimdbId('');
  };
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,109 @@
...
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(event) => {
+          setTitle(event); // why this way
+        }}
```
meforoma, line 59: your question is spot-on here.
that might be a default behaviour for `onChange`
so it does work as-is, and you can use a shorthand syntax
```suggestion
        onChange={setTitle}
```

yet of course it's better to remain explicit
```
(event) => setState(event.target.value)
```

meforoma (approved):
fully functional 👍🏼 
good job

PR #1241, opened 2023-02-02
Enhance the add movie form and its validation

mykyta-krainik:
[DEMO LINK](https://mykyta-krainik.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
+import { FC, useState } from 'react';
+
+import { Movie } from './types/Movie';
+
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
 
-export const App = () => {
+export const App: FC = () => {
+  const [movies, setMovies] = useState<Movie[]>([...moviesFromServer]);
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie
+          onAdd={(movie) => setMovies([...movies, movie])}
+        />
```
polosanya, line 21: Better save this function into some variable and pass it here

polosanya (changes requested):
The errors are shown, but I still can add a film. I think it's better to show another error message if links don't match the needed format
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,118 @@
...
+
+  const [isImgUrlValid, setIsImgUrlValid] = useState(true);
+  const [isImdbUrlValid, setIsImdbUrlValid] = useState(true);
+
+  const isButtonDisabled = !title
+    || !imgUrl || !isImgUrlValid
+    || !imdbUrl || !isImdbUrlValid
+    || !imdbId;
+
+  function handleSubmit(event: FormEvent<HTMLFormElement>): void {
```
agniya-i, line 32: better to use arrow function

agniya-i (approved):

PR #1244, opened 2023-02-03
solution

Malva37:
[DEMO LINK](https://Malva37.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,112 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie)=> void,
+};
+
+export const NewMovie:React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const onReset = () => {
+    setCount(current => current + 1);
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,112 @@
...
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
         required
+        onChange={(e) => {
```
anna-inozemets, line 67: Maybe it is better to use another name for variable in parameters instead of 'e'. (I guess it is better to use smth like "newTitle". I also, guess that it ill be good to rename it in all places.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,112 @@
...
+export const NewMovie:React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const onReset = () => {
+    setCount(current => current + 1);
```
anna-inozemets, line 18: Maybe it is better to rename current to currentCount, for better understanding

anna-inozemets (approved):
Good and clear solution!)

src/components/NewMovie/NewMovie.tsx
Malva37: nice advice, thanks

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,112 @@
...
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
         required
+        onChange={(newTitle) => {
+          setTitle(newTitle);
+        }}
```
denys-cheporniuk, line 69: you don't have to create arrow function here, pass only setTitle method, fix in all places

denys-cheporniuk (changes requested):

src/components/NewMovie/NewMovie.tsx
denys-cheporniuk: or, even better, create a separate function and use it here

denys-cheporniuk (approved):

PR #1247, opened 2023-02-03
solution

igor-kashchenko:
[DEMO LINK](https://igor-kashchenko.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,29 @@
-import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+
+import './App.scss';
+
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = (movie: Movie) => {
+    setMovies([
+      ...movies,
+      movie,
+    ]);
+  };
+
   return (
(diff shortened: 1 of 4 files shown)


OlhaYevstifieieva (approved):
Well done!

denys-cheporniuk (approved):

src/App.tsx
```diff
@@ -1,16 +1,29 @@
...
+import './App.scss';
+
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = (movie: Movie) => {
+    setMovies([
```
Anna-Tkachenko, line 14: use functional setState if you need to set new data according to the current one

Anna-Tkachenko (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,90 @@
...
+  const handleSubmit = (event: FormEvent) => {
+    event.preventDefault();
+
+    onAdd({
+      title,
+      description,
+      imgUrl,
+      imdbId,
+      imdbUrl,
+    });
```
Anna-Tkachenko, line 38: will be better to store the main data in separate variable initially
and then pass those variable to the function
instead of creating an object while calling the function in the argument

Anna-Tkachenko (approved):

PR #1258, opened 2023-02-08
solution

Oleksandr-Maly:
[DEMO LINK](https://oleksandr-maly.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -7,8 +7,6 @@ type Props = {
 };
 
 export const NewMovie:React.FC<Props> = ({ onAdd }) => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +92,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            {... { disabled: !conditionToSubmit }}
```
denys-cheporniuk, line 95: wow, you don't need spread operator here

denys-cheporniuk (approved):

src/components/NewMovie/NewMovie.tsx
Oleksandr-Maly: > wow, you don't need spread operator here

@denys-cheporniuk тоді свариться

PR #1268, opened 2023-02-10
solution

miamarichka:
[DEMO LINK](https://miamarichka.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,33 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+  const addNewMovie = ({
+    title, description, imgUrl, imdbUrl, imdbId,
+  }: Movie) => {
+    const newMovie = {
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    };
+
+    setMovies([...movies, newMovie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addNewMovie} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,82 @@
 import { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie:Movie) => void,
+};
+
+export const NewMovie: React.FC <Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isRequired = title && imgUrl && imdbUrl && imdbId;
+  const disabled = !isRequired;
+
+  const reset = () => {
+    setTitle('');
+    setDescription('');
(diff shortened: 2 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,82 @@
...
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie:Movie) => void,
+};
+
+export const NewMovie: React.FC <Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
```
Anastasiia-Svintsova, line 11: Remove comments

src/components/TextField/TextField.tsx
```diff
@@ -18,14 +18,19 @@ export const TextField: React.FC<Props> = ({
   value,
   label = name,
   required = false,
-  onChange = () => {},
+  onChange = () => { },
```
Anastasiia-Svintsova, line 21: Why do you need this empty onChange function? Just made this item required

src/components/TextField/TextField.tsx
```diff
@@ -36,14 +41,18 @@ export const TextField: React.FC<Props> = ({
...
           })}
           type="text"
           placeholder={`Enter ${label}`}
-          value={value}
-          onChange={event => onChange(event.target.value)}
+          value={textareaValue}
+          onChange={event => {
+            setTextareaValue(event.target.value);
+            onChange((event.target.value));
+          }}
```
Anastasiia-Svintsova, line 55: Move it to a function

src/components/TextField/TextField.tsx
```diff
@@ -18,14 +18,19 @@ export const TextField: React.FC<Props> = ({
...
 
   // To show errors only if the field was touched (onBlur)
   const [touched, setToched] = useState(false);
-  const hasError = touched && required && !value;
+  const [textareaValue, setTextareaValue] = useState(value);
+  const hasError = touched && required && !textareaValue;
+
+  const isNeedPattern = name.toLowerCase().includes('url');
+  // eslint-disable-next-line max-len, no-useless-escape
+  const pattern = '/^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A - Za - z0 - 9. -] +| (?: www\.| [-;:&=+$, \w]+@)[A - Za - z0 - 9. -] +) ((?: \/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/';
```
Anastasiia-Svintsova, line 33: Seems that it doesn't work. Check it, please

Anastasiia-Svintsova (changes requested):
- I can add a movie with empty fields. Fix it, please 

https://user-images.githubusercontent.com/99403198/218257449-cd4b9259-18eb-42d9-b16e-d688b22f5f37.mov

src/components/TextField/TextField.tsx
miamarichka: It was optional task, I`ve tried to make it but I really dont know how to do. I deleted it. I searched hoe to make it, but didnt find any good answer. So maybe you can recommend any tutorial, book or show how it should be correct?

src/App.tsx
```diff
@@ -1,16 +1,33 @@
...
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+  const addNewMovie = ({
+    title, description, imgUrl, imdbUrl, imdbId,
+  }: Movie) => {
```
Anastasiia-Svintsova, line 12: It's better to destructure the value inside the function
```suggestion
  const addNewMovie = (movie: Movie) => {
    const { title, description, imgUrl, imdbUrl, imdbId } = movie;
```

Anastasiia-Svintsova (approved):

src/components/TextField/TextField.tsx
Anastasiia-Svintsova: You could use a test() method for example. (Can read [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) about it)
```
const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
cconst url = "some URL or text"
```

and then if you do this `pattern.test(url)` the output will be true (if it matches) or false (if not)

PR #1274, opened 2023-02-12
solution

imariash86:
[DEMO LINK](https://imariash86.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,22 @@
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
 
 export const App = () => {
+  const [visibleMovies, setVisibleMovies] = useState([...moviesFromServer]);
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={visibleMovies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie) => {
+          setVisibleMovies([...visibleMovies, movie]);
+        }}
+        />
       </div>
     </div>
   );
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,94 @@
...
+
+  const isValid = inputTitle && inputImage && inputUrl && inputId;
+
+  const movie = {
+    title: inputTitle,
+    description: inputDescription,
+    imgUrl: inputImage,
+    imdbUrl: inputUrl,
+    imdbId: inputId,
+  };
```
oleh-holovnykh, line 32: const movie is not needed until submit
maybe it would be better to declare a movie in `handleSubmit`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,94 @@
...
 
   return (
-    <form className="NewMovie" key={count}>
+    <form
+      className="NewMovie"
+      key={count}
+      onSubmit={(event) => {
+        handleSubmit(event);
+      }}
```
oleh-holovnykh, line 51: no extra wrap needed

```suggestion
      onSubmit={handleSubmit}
```

oleh-holovnykh (changes requested):
Good job!👍
There are few suggestions

imariash86:
> Good job!👍 There are few suggestions

thanks a lot for your suggestions, it's really reasonable:)

oleh-holovnykh (approved):
🔥

src/App.tsx
```diff
@@ -1,16 +1,28 @@
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [visibleMovies, setVisibleMovies] = useState([...moviesFromServer]);
```
Anna-Tkachenko, line 9: you shouldn't make additional array copy while setting array as the initial value, it's redundant here
there is no cases, when we can mutate the properties in state (each useState call create a copy inside)

Anna-Tkachenko: and check variable name, will be better to name it simpler, just `movies`

Anna-Tkachenko (approved):

PR #1304, opened 2023-02-20
movies list add form completed

MykolaKaradzha:
https://mykolakaradzha.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,98 @@
-import { useState } from 'react';
+import React, { FormEvent, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState<number>(0);
+  const [title, setTitle] = useState<string>('');
+  const [description, setDescription] = useState<string>('');
+  const [imgUrl, setImgUrl] = useState<string>('');
+  const [imdbUrl, setImdbUrl] = useState<string>('');
+  const [imdbId, setImdbId] = useState<string>('');
+  const [validationError, setValidationError] = useState<boolean>(false);
+
+  const isDisabled = !(!!title && !!imgUrl && !!imdbUrl && !!imdbId);
+  const urlValidation = (url: string) => {
+    // eslint-disable-next-line max-len
+    const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
+
+    return url.match(pattern);
+  };
+
+  const movie = {
+    title,
+    description,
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,98 @@
...
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState<number>(0);
+  const [title, setTitle] = useState<string>('');
+  const [description, setDescription] = useState<string>('');
+  const [imgUrl, setImgUrl] = useState<string>('');
+  const [imdbUrl, setImdbUrl] = useState<string>('');
+  const [imdbId, setImdbId] = useState<string>('');
+  const [validationError, setValidationError] = useState<boolean>(false);
```
etojeDenys, line 18: we can combine it in one useState({
count: 0,
title: '',
...
})

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,98 @@
...
+
+  const isDisabled = !(!!title && !!imgUrl && !!imdbUrl && !!imdbId);
+  const urlValidation = (url: string) => {
+    // eslint-disable-next-line max-len
+    const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
+
+    return url.match(pattern);
+  };
```
etojeDenys, line 26: actually, we can move this function out of the component

etojeDenys (changes requested):
so, create the useState(inititalFields)
initialFields would be the object with default fields data

also you can create one `handleChange` function and change state like this
```
setFields({...fields, [e.target.name]: e.target.vaue})
```

etojeDenys:
if you have some questions you always can ask them in the slack chat

src/components/NewMovie/NewMovie.tsx
MykolaKaradzha: Thanks for remarks. I implemented your suggestions, though not sure, that I did right. Its more bulky now).

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,132 @@
...
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [initialFields, setFields] = useState<InitialFields>({
+    count: 0,
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+    validationError: false,
```
IvanFesenko, line 36: You can create separate const for it, like 
```
const EMPTY_STATE= {
    count: 0,
    title: '',
    description: '',
    imgUrl: '',
    imdbUrl: '',
    imdbId: '',
    validationError: false,
}
```
and use it here useState<InitialFields>(EMPTY_STATE) and below

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,132 @@
...
+    onAdd(movie);
+    setFields(state => ({
+      count: state.count + 1,
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+      validationError: false,
+    }));
```
IvanFesenko, line 84: And here
```suggestion
    setFields(state => ({
      ...EMPTY_STATE,
      count: state.count + 1,
    }));
```

IvanFesenko (approved):

src/components/NewMovie/NewMovie.tsx
MykolaKaradzha: That`s nice idea, thank you).

PR #1324, opened 2023-03-12
React_movies-list-add-form

SerhiiVovchenko:
[DEMO LINK](https://SerhiiVovchenko.github.io/react_movies-list-add-form/)


agniya-i (changes requested):
When some of the links not valid, the form resets with this errors, looks strange from user perspective, better not to reset form and show a validation error for those fields


https://user-images.githubusercontent.com/32853034/224555153-f258e502-ba9d-40f7-9c16-e278440dada7.mov

src/components/TextField/TextField.tsx
```diff
@@ -49,7 +51,7 @@ export const TextField: React.FC<Props> = ({
       </div>
 
       {hasError && (
-        <p className="help is-danger">{`${label} is required`}</p>
+        <p className="help is-danger">{textError ? label + textError : `${label} is required`}</p>
```
agniya-i, line 54:
```suggestion
        <p className="help is-danger">{textError && `${label + textError}` </p>
```
pay attention that you always have some value in `textError`, so this check can be redutant

agniya-i (approved):

PR #1339, opened 2023-03-18
Add New Field Form 1

mdm-21:
[DEMO LINK](https://mdm-21.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,88 @@
...
+
+  const hasValidSection = title && imgUrl && imdbUrl && imdbId;
+
+  const addNewMovie: Movie = {
+    title,
+    description,
+    imgUrl,
+    imdbUrl,
+    imdbId,
+  };
```
ivbutsykin, line 25: Move it inside `handleSubmit` function, no need to create it here.
Also rename it in `newMovie` (because `addNewMovie` sounds like function)

ivbutsykin (approved):
I can add new film when urls not valid, could you try to fix it?
Screenshot 2023-03-18 at 14 20 33


Overall, all seems good, so I approved it.

PR #1341, opened 2023-03-18
solution

vlad-trundenko:
[DEMO LINK](https://vlad-trundenko.github.io/react_movies-list-add-form/)


OleziO (changes requested):
image
Add trim to all fields.

agniya-i (approved):

PR #1346, opened 2023-03-24
Solution

ifuride:
[DEMO LINK](https://ifuride.github.io/react_movies-list-add-form/)


AlisaVlasova (changes requested):
I can add empty movie, trim inputs value 
image

OleziO (approved):

PR #1348, opened 2023-03-25
the task was solved

YevheniiHura:
https://YevheniiHura.github.io/react_movies-list-add-form/


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,90 @@
...
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
VitaliyBondarenko1982, line 16:
```suggestion
  const [inputs, setInputs] = useState({
    title: '',
    description: '',
    imgUrl: '',
    imdbUrl: '',
    imdbId: ''
  })
  
  const onChange = (e: React.ChangeEvent<HtmlInputElement>) => {
     const [name, value] = event.target;
     
     setInputs(prevInputs => { ...prevInputs, [name]: value })
  }
```
we can combine all inputs states in one state and use input name to change some field.
will be something like that

VitaliyBondarenko1982 (changes requested):
check and push all files
can't see MoviesList and TextField components
Screenshot from 2023-03-25 18-29-46

src/components/NewMovie/NewMovie.tsx
YevheniiHura: done

etojeDenys (approved):
LGTM

PR #1350, opened 2023-03-26
solution

Roman-Malysh:
https://Roman-Malysh.github.io/react_movies-list-add-form/

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -13,3 +13,8 @@
   padding: 20px;
   border-left: 1px solid lightgray;
 }
+
+iframe {
+  pointer-events: none;
+  z-index: 1;
+}
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,97 @@
-import { useState } from 'react';
+import { FormEvent, useState } from 'react';
+import { Movie } from '../../types/Movie';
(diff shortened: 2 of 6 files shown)


src/App.scss
```diff
@@ -13,3 +13,8 @@
   padding: 20px;
   border-left: 1px solid lightgray;
 }
+
+iframe {
+  pointer-events: none;
+  z-index: 1;
+}
```
tetiana-ivanchenko, line 20: why do you need this styles

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,97 @@
...
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
```
tetiana-ivanchenko, line 11: remove comments

tetiana-ivanchenko (changes requested):

src/App.scss
Roman-Malysh: I face a bug sometimes when iframe with a size of all screen appears and prevents all clicks on page

PR #1352, opened 2023-03-27
Solution

VadymTsyndra:
https://VadymTsyndra.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -14,20 +14,17 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   const [imdbUrl, setImdbUrl] = useState('');
   const [imdbId, setImdbId] = useState('');
 
-  const areValuesEntered = title !== ''
-    && imgUrl !== ''
-    && imdbUrl !== ''
-    && imdbId !== '';
+  const areValuesEntered = title && imgUrl && imdbUrl && imdbId;
 
   const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
     event.preventDefault();
 
     onAdd({
-      title,
-      description,
-      imgUrl,
-      imdbUrl,
-      imdbId,
+      title: title.trim(),
+      description: description.trim(),
+      imgUrl: imgUrl.trim(),
+      imdbUrl: imdbUrl.trim(),
+      imdbId: imdbId.trim(),
     });
 
     setTitle('');


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,88 @@
...
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [valueTitle, setValue] = useState('');
+  const [valueDescription, setValueDescription] = useState('');
```
polosanya, line 12:
```suggestion
  const [title, setTitle] = useState('');
  const [description, setDescription] = useState('');
```
Set function should be named according to the noun variable. Also, do you really need to duplicate "value" everywhere? I think it's redundant. Please fix all cases

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,88 @@
...
+  const [valueImageUrl, setValueImageUrl] = useState('');
+  const [valueImdbUrl, setValueImdbUrl] = useState('');
+  const [valueImdbId, setValueImdbId] = useState('');
+
+  const checkButton
+    = valueTitle !== ''
+    && valueImageUrl !== ''
+    && valueImdbUrl !== ''
+    && valueImdbId !== '';
```
polosanya, line 21:
```suggestion
  const areValuesEntered = valueTitle !== ''
    && valueImageUrl !== ''
    && valueImdbUrl !== ''
    && valueImdbId !== '';
```
Fix formatting. Also, it's better to name boolean variables according to these rules, so it's clear that it is a boolean value
https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,88 @@
...
+        });
+
+        setValue('');
+        setValueDescription('');
+        setValueImageUrl('');
+        setValueImdbUrl('');
+        setValueImdbId('');
+
+        setCount(current => current + 1);
+      }}
```
polosanya, line 45: Don't store it here. Create a separate function to handle submit

polosanya (changes requested):
Almost done! Let's improve your code a bit)

OleziO (changes requested):
image
Add string trim to avoid empty fields.

anstsot (approved):
Great job, approved!

PR #1353, opened 2023-03-28
addFilm form

Mykola-Busarhin:
[DEMO LINK](https://Mykola-Busarhin.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,87 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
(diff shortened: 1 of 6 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +90,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={Boolean(!title || !imgUrl || !imdbId || !imdbUrl)}
```
tetiana-ivanchenko, line 93: use trim, because now I can add film with just spaces in all fields

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #1355, opened 2023-03-29
movies add react

valeriy-mikheyev:
[DEMO LINK](https://valeriy-mikheyev.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,90 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const resetForm = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,90 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
IvanFesenko, line 17: You can use state with object with all this fields
```
const EMPTY_MOVIE = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: '',
}

const [movie, setMovie] = useState(EMPTY_MOVIE)
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,90 @@
...
+
+  const resetForm = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
```
IvanFesenko, line 24: And here set by one call 
```suggestion
  setMovie(EMPTY_MOVIE)
```

IvanFesenko (changes requested):
Check this case
https://user-images.githubusercontent.com/61904995/228528769-72903937-023b-4bfb-ab3a-479a212eed4a.mp4

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,106 @@
...
+
+    onAdd(movie);
+
+    setCount(prev => prev + 1);
+
+    resetForm();
+  };
+
+  const validateUrl = (url: string) => {
+    return pattern.test(url);
```
slavanosatskyi, line 48: I'm not sure that it works as expected as i can still add movie with non-valid urls
Screenshot 2023-03-29 at 18 13 48. But this is optional requirement, so you can skip it for now

slavanosatskyi (approved):

PR #1363, opened 2023-03-30
Added React Movies list add form

ctk12:
[DEMO LINK](https://ctk12.github.io/react_movies-list-add-form/)


qoa11a (changes requested):
The demo link doesn't work. Let's fix it and re-request a review.

PR #1389, opened 2023-04-05
solution

IhorChaikovskyi:
[DEMO LINK](https://IhorChaikovskyi.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,123 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (currentState: any) => void,
+};
+
+export const NewMovie:React.FC<Props> = ({
+  onAdd,
+}) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [objMovie, setObjMovie] = useState({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+
+  const areAllFieldsFilledIn = (obj: Movie) => {
+    if (
+      obj.title === ''
+      || obj.imdbId === ''
+      || obj.imdbUrl === ''
+      || obj.imgUrl === '') {
+      return true;
+    }
+
+    return false;
+  };
 
   return (
-    <form className="NewMovie" key={count}>
-      <h2 className="title">Add a movie</h2>
+    <form
+      className="NewMovie"
+      key={count}
+      onSubmit={(event) => {
+        event.preventDefault();
+        if (areAllFieldsFilledIn(objMovie)) {
+          return;
+        }
 
+        setCount(count + 1);
+
+        setObjMovie({
+          title: '',
+          description: '',
+          imgUrl: '',
+          imdbUrl: '',
+          imdbId: '',
+        });
+
+        onAdd(objMovie);
+      }}
+    >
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,123 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (currentState: any) => void,
```
anstsot, line 6: set proper typing
```suggestion
  onAdd: (movie: Movie) => void,
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,123 @@
...
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+
+  const areAllFieldsFilledIn = (obj: Movie) => {
+    if (
+      obj.title === ''
```
anstsot, line 25: also check for empty spaces. try yo use `.trim()` function

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,123 @@
...
+        setObjMovie({
+          title: '',
+          description: '',
+          imgUrl: '',
+          imdbUrl: '',
+          imdbId: '',
+        });
+
+        onAdd(objMovie);
+      }}
```
anstsot, line 56: create separate function `onSubmit` and move the code there
```suggestion
      onSubmit={onSubmit}
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,123 @@
...
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={objMovie.title}
+        onChange={titleOnInput => setObjMovie((currentState: Movie) => {
+          return {
+            ...currentState,
+            title: titleOnInput,
+          };
+        })}
```
anstsot, line 68: it can be simplified
```suggestion
        onChange={title => setObjMovie((prevState) => ({ ...prevState, title }))}
```

anstsot (changes requested):
It’s almost done! Let’s improve a little.
And check the demo link, please. It's not working

src/components/NewMovie/NewMovie.tsx
IhorChaikovskyi: Thank you.This is good advice!

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,111 @@
...
+    if (
+      obj.title.trim() === ''
+      || obj.imdbId.trim() === ''
+      || obj.imdbUrl.trim() === ''
+      || obj.imgUrl.trim() === '') {
+      return true;
+    }
+
+    return false;
+  };
```
sTorba24, line 33:
```suggestion
  const сheckEmptyFields = (obj: Movie) => {
    if (
      obj.title.trim()
      || obj.imdbId.trim()
      || obj.imdbUrl.trim()
      || obj.imgUrl.trim()) {
      return true;
    }

    return false;
  };
```

sTorba24: `areAllFieldsFilledIn` and this function returning `true` when one of fields has no value
Seems can be named better

sTorba24 (changes requested):
Almost good! lets improve a little

src/components/NewMovie/NewMovie.tsx
IhorChaikovskyi: Yes, I agree with that

tetiana-ivanchenko (approved):

PR #1398, opened 2023-04-10
Solution

GUSILLUS:
- [DEMO LINK](https://GUSILLUS.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,94 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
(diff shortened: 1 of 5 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -18,14 +18,18 @@ export const TextField: React.FC<Props> = ({
   value,
   label = name,
   required = false,
-  onChange = () => {},
+  onChange,
 }) => {
-  // generage a unique id once on component load
+  // generage a unique id once on component load0
```
Viktor-Kost, line 23: remove comments
```suggestion
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +97,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!isAddButtonDisabled}
```
Viktor-Kost, line 100: update this condition to get button state to be clearer. 
Because in this case `!isAddButtonDisabled` it equal to `isAddButtonAbled`, so how `disabled` can be set when button is abled 
```suggestion
            disabled={isAddButtonDisabled}
```

Viktor-Kost (changes requested):

PR #1408, opened 2023-04-14
solution

Yeliseiev:
[DEMO LINK](https://Yeliseiev.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,17 +9,13 @@ export const App: React.FC = () => {
   const [movies, setMovies] = useState(moviesFromServer);
   const [count, setCount] = useState(0);
 
-  const increseCount = () => {
-    setCount(currentCount => currentCount + 1);
-  };
-
   const addNewMovie = (newMovie: Movie) => {
     setMovies(currentMovies => ([
       ...currentMovies,
       newMovie,
     ]));
 
-    increseCount();
+    setCount(currentCount => currentCount + 1);
   };
 
   return (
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,16 +1,34 @@
...
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+  const [count, setCount] = useState(0);
+
+  const increseCount = () => {
```
mykhalenych, line 12: The 'increseCount' function has a typo. Consider renaming it to 'increaseCount'

src/App.tsx
```diff
@@ -1,16 +1,34 @@
...
+    setCount(currentCount => currentCount + 1);
+  };
+
+  const addNewMovie = (newMovie: Movie) => {
+    setMovies(currentMovies => ([
+      ...currentMovies,
+      newMovie,
+    ]));
+
+    increseCount();
```
mykhalenych, line 22: Instead of using 'increaseCount' function, you can directly update the 'movieCount' state inside 'addNewMovie' function.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,85 @@
...
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = (
+  { onAdd },
+) => {
+  const [count, setCount] = useState(0);
```
mykhalenych, line 13: why do you setCount twice?
use prop count={count}

Yeliseiev: The key in the form does not pass the tests if I take it from the props, so count in NewMovie.tsx is only for tests.

mykhalenych (approved):

PR #1409, opened 2023-04-14
add-form-react

fenderukr:
https://fenderukr.github.io/react_movies-list-add-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,25 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,16 +1,25 @@
...
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie
+          onAdd={(movie: Movie) => setMovies(
+            (prevState) => [...prevState, movie],
+          )}
+        />
+        {/* onAdd = propsName, we pass a function to it that sets a new state with all previous movies + a new one from a filled form */}
```
loralevitska, line 22: Remove comments. Fix it everywhere

src/App.tsx
```diff
@@ -1,16 +1,25 @@
...
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie
+          onAdd={(movie: Movie) => setMovies(
+            (prevState) => [...prevState, movie],
+          )}
```
loralevitska, line 20: Move to separate function  (transfer the function to a variable, and here only write the name of the function)

loralevitska (changes requested):
Good job! There are things that should be fixed:
- i can add empty card (only spaces). Fix it with string.trim()

src/App.tsx
fenderukr: > 

yes, but this is how mate's example works. I can also create a card by entering just spaces there

polosanya (approved):
Good job!

PR #1416, opened 2023-04-20
implemented movie add form

Esteban4769:
[DEMO LINK](https://esteban4769.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,10 +1,59 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie:React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [movieFields, setMovieFields] = useState<Movie>({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +104,8 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!fieldIsFilled}
+            onClick={handleClick}
```
PavloReutskiy, line 108: you should move it to the form tag according to checklist and use onSubmit instead of onClick

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,104 @@
...
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie:React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
```
SerhiiKirik, line 10: remove extra count
```suggestion
```

SerhiiKirik (changes requested):
Well done 🔥 
I can add a card just with spaces
Let's validate your fields
image

src/components/NewMovie/NewMovie.tsx
Esteban4769: `count` is  for new form identification and its in task by default. I rename it for clarity

SerhiiKirik: But you don't need a key property into form
https://legacy.reactjs.org/docs/lists-and-keys.html#keys

SerhiiKirik (approved):
LGTM 🔥

PR #1424, opened 2023-04-25
Solution

kate-matviiv:
https://kate-matviiv.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -41,6 +41,11 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   };
 
   const isFieldsValid = checkingFields();
+  const isValidImdbId = (id: string): boolean => {
+    const imdbIdPattern = /^tt\d{7}$/;
+
+    return imdbIdPattern.test(id);
+  };
 
   const isTitleValid = newMovie.title.trim().length > 0;
 
@@ -107,7 +112,7 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
         label="Imdb ID"
         value={newMovie.imdbId}
         onChange={handler}
-        checkURL={validationURL}
+        checkURL={isValidImdbId}
         required
       />
 


sTorba24 (changes requested):
Almost done! 

i'm not able to add a description, fix it please

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,108 @@
...
+  };
+
+  const clearForm = () => {
+    setNewMovie({
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
```
OleziO, line 33: Create a default object from this and set it.

OleziO (changes requested):
Add field validation to avoid this case.
image

OleziO (changes requested):
Good Job! Disable the button when validation does not pass to avoid adding an empty item.
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +114,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={checkingFields() || !isTitleValid}
```
Anastasiia-Svintsova, line 117: don't call the function inside JSX. Move the result of a function to a variable and use it

Anastasiia-Svintsova (changes requested):
- IMDb id is a required field. But now I can add a movie if it's empty

https://user-images.githubusercontent.com/99403198/234841332-c8901d91-dd2d-4b2a-af91-5d2725b7ea3d.mov

src/components/TextField/TextField.tsx
```diff
@@ -41,16 +41,26 @@ export const TextField: React.FC<Props> = ({
...
         <p className="help is-danger">{`${label} is required`}</p>
       )}
+
+      {
+        value && !checkURL(value) && (
+          <p className="help is-danger">{`${label} isn't correct`}</p>
+        )
+      }
+
     </div>
```
polosanya, line 64:
```suggestion
      {value && !checkURL(value) && (
        <p className="help is-danger">
          {`${label} isn't correct`}
        </p>
      )}
    </div>
```

polosanya (approved):
Well done!

PR #1427, opened 2023-04-26
solution

makandriival:
[DEMO LINK](https://makandriival.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const onAdd = (movie: Movie) => {
+    setMovies([...movies, movie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={onAdd} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,110 @@
-import { useState } from 'react';
+import { FormEvent, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+interface Props {
+  onAdd: (movie: Movie) => void,
+}
+
+export const NewMovie = ({ onAdd }: Props) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const handleOnChange = (string: string, label: string) => {
+    switch (label) {
+      case 'Title':
+        setTitle(string);
+        break;
+      case 'Description':
+        setDescription(string);
+        break;
+      case 'Image URL':
+        setImgUrl(string);
+        break;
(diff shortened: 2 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,110 @@
...
+import { FormEvent, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+interface Props {
+  onAdd: (movie: Movie) => void,
+}
+
+export const NewMovie = ({ onAdd }: Props) => {
```
igorkruz, line 9: export const NewMovie: React.FC<Props> = ({ onAdd }) => {}

https://mate-academy.github.io/fe-program/js/extra/react-typescript
import

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,110 @@
...
+    const movie = {
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    };
+
+    onAdd(movie);
+    setCount(count + 1);
```
igorkruz, line 59: I think here you need to refer to the previous state because it is important to get the correct count

`setCount(prevCount => prevCount + 1);`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +113,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!title || !imgUrl || !imdbUrl || !imdbId}
```
igorkruz, line 116: You can move this into a separate variable.

igorkruz (changes requested):
Good job 👍 
I think with these suggested changes your code will be better.

src/components/NewMovie/NewMovie.tsx
makandriival: nice idea=)

igorkruz (approved):
Good job 👍

src/components/TextField/TextField.tsx
```diff
@@ -43,7 +43,7 @@ export const TextField: React.FC<Props> = ({
           type="text"
           placeholder={`Enter ${label}`}
           value={value}
-          onChange={event => onChange(event.target.value)}
+          onChange={event => onChange(event.target.value, label)}
```
liliia-livandovskaya, line 46: Check if 'label' is defined before passing it to the 'onChange' callback: event => onChange(event.target.value, label || '')

src/App.tsx
```diff
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
```
liliia-livandovskaya, line 9: It is a good practice to type the initial state when using useState. Consider typing the initial state like this: useState<Movie[]>(moviesFromServer).

liliia-livandovskaya (changes requested):
1. Add `onBlur` to check that title, imgUrl, imdbUrl, imdbId are entered when an input looses focus  and show an error and a red border if needed;
2. Add validation to check fields input 
Screenshot 2023-05-02 at 15 59 17

src/components/TextField/TextField.tsx
makandriival: why do i need to do this? Label is always defined, as far as i can tell

src/components/TextField/TextField.tsx
```diff
@@ -6,7 +6,7 @@ type Props = {
   value: string,
   label?: string,
   required?: boolean,
-  onChange?: (newValue: string) => void,
+  onChange?: (newValue: string, label: string) => void,
```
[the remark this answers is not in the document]
makandriival: again, i cant get it =) my bad if i'm being dumb. When i do this changes i get an TS error, help pls

src/components/TextField/TextField.tsx
liliia-livandovskaya: Adding the label as a second argument to the onChange function is not mandatory, but it is recommended as a best practice.

The addition of label as a second argument to the onChange function is to provide additional context or metadata to the function handling the change event.

By passing label as a second parameter, you can differentiate between multiple input fields that use the same change event handler function. This makes it easier to handle form inputs, especially when there are multiple fields and you want to update the state or perform certain actions based on which input field triggered the event.

liliia-livandovskaya (approved):

PR #1428, opened 2023-04-26
created react movie list add form

igorkruz:
[DEMO LINK](https://igorkruz.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,7 +8,7 @@ import moviesFromServer from './api/movies.json';
 export const App: React.FC = () => {
   const [visibleMovies, setVisibleMovies] = useState(moviesFromServer);
 
-  const onAdd = (newMovie: Movie) => {
+  const handleAdd = (newMovie: Movie) => {
     setVisibleMovies((currentMovies => [...currentMovies, newMovie]));
   };
 
@@ -18,7 +18,7 @@ export const App: React.FC = () => {
         <MoviesList movies={visibleMovies} />
       </div>
       <div className="sidebar">
-        <NewMovie onAdd={onAdd} />
+        <NewMovie onAdd={handleAdd} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -15,7 +15,18 @@ export const NewMovie:FC<Props> = ({ onAdd }) => {
   const [imdbUrl, setImdbUrl] = useState('');
   const [imdbId, setImdbId] = useState('');
 
-  const isValidForm = title && imgUrl && imdbUrl && imdbId;
+  const isValidForm = title.trim()
+    && imgUrl.trim()
+    && imdbUrl.trim()
+    && imdbId.trim();
+
+  const handleReset = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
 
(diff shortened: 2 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,89 @@
...
+
+export const NewMovie:FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
```
PashaMalyshkin, line 17: You can use only one useState with object here and set its value in onChange, it looks more clear and simple i think, but you can use your variant, it's just suggestion :)

```suggestion
export const NewMovie:FC<Props> = ({ onAdd }) => {
  const [count, setCount] = useState(0);

  const [newMovie, setMovie] = useState({
    title: '',
    description: '',
    imgUrl: '',
    imdbUrl: '',
    imdbId: '',
  });

```

PashaMalyshkin (approved):

src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+import { Movie } from './types/Movie';
 import moviesFromServer from './api/movies.json';
 
-export const App = () => {
+export const App: React.FC = () => {
+  const [visibleMovies, setVisibleMovies] = useState(moviesFromServer);
+
+  const onAdd = (newMovie: Movie) => {
```
SerhiiKirik, line 11: In the upper component - `handle`, in the down - `on`
```suggestion
  const handleAdd = (newMovie: Movie) => {
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,89 @@
...
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (newMovie: Movie) => void
+}
+
+export const NewMovie:FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
```
SerhiiKirik, line 10: remove extra state
```suggestion
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,89 @@
...
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    };
+
+    onAdd(newMovie);
+
+    setTitle('');
```
SerhiiKirik, line 33: make a separate `func` - `handleReset`

SerhiiKirik (changes requested):
Well done 🔥 
I can add a card just with spaces
Let's validate your fields
image

src/components/NewMovie/NewMovie.tsx
igorkruz: if I remove this count state I will need to remove it from the form I used it in the form like a key property and  if I remove the key from the form I got an error after submitting a form 
an error like this
Screenshot 2023-05-02 083246

SerhiiKirik (approved):
Great 🔥

PR #1435, opened 2023-04-27
completed movie list add form

VenomMB:
[DEMO LINK](https://VenomMB.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,27 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+import { Movie } from './types/Movie';
 import moviesFromServer from './api/movies.json';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const onAddFilm = (movie: Movie) => {
+    setMovies(previousMovies => ([
+      ...previousMovies,
+      movie,
+    ]));
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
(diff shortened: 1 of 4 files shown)


vksazonov (approved):
GJ!!!

src/App.tsx
```diff
@@ -1,16 +1,27 @@
...
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+import { Movie } from './types/Movie';
 import moviesFromServer from './api/movies.json';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const onAddFilm = (movie: Movie) => {
```
SerhiiKirik, line 11:
```suggestion
  const handleAdd = (movie: Movie) => {
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,96 @@
...
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd?: (movie: Movie) => void;
+}
+
+export const NewMovie: FC<Props> = ({ onAdd = () => {} }) => {
+  const [count, setCount] = useState(0);
```
SerhiiKirik, line 11: remove extra state
```suggestion
```

SerhiiKirik (changes requested):
Well done 🔥 
i can add card just with spaces
Let's validate your fields
image

src/components/NewMovie/NewMovie.tsx
VenomMB: I can't remove this state because I use it for the form key. And I had this state by default.

SerhiiKirik (approved):
Great 🔥 
It's a bad way not to allow them to user put a space

VenomMB:
> Great 🔥 It's a bad way not to allow them to user put a space

Why, the  user can use space, but not at the start
image

PR #1438, opened 2023-04-27
React Movies list add form v.0.1 - Solution

NedobezhkinOleksandr:
https://NedobezhkinOleksandr.github.io/react_movies-list-add-form/


tetiana-ivanchenko (changes requested):
Now I can add film with just spaces in all fields

anstsot (approved):
Great job, approved!

PR #1441, opened 2023-04-28
commit

anastasia-radchencko:
https://anastasia-radchencko.github.io/react_movies-list-add-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,34 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [visibleMovie, setVisibleMovie] = useState([...moviesFromServer]);
+
+  const addNewMovie = (
+    movie: Movie,
+  ) => {
+    const newMovie = {
+      title: movie.title,
+      description: movie.description,
+      imgUrl: movie.imgUrl,
+      imdbUrl: movie.imdbUrl,
+      imdbId: movie.imdbId,
+    };
+
+    setVisibleMovie(prevValue => [...prevValue, newMovie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={visibleMovie} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addNewMovie} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,89 @@
-import { useState } from 'react';
+import { FormEvent, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd(
+    movie: Movie
+  ): void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const handleFormSubmit = (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (title && imgUrl && imdbUrl && imdbId) {
+      onAdd({
+        title,
+        description,
+        imgUrl,
+        imdbUrl,
+        imdbId,
+      });
+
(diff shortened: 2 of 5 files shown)


src/App.tsx
```diff
@@ -1,16 +1,34 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [visibleMovie, setVisibleMovie] = useState([...moviesFromServer]);
```
ivbutsykin, line 9: No need to create another copy of array here:
```suggestion
  const [visibleMovie, setVisibleMovie] = useState(moviesFromServer);
```

src/App.tsx
```diff
@@ -1,16 +1,34 @@
...
+  const addNewMovie = (
+    movie: Movie,
+  ) => {
+    const newMovie = {
+      title: movie.title,
+      description: movie.description,
+      imgUrl: movie.imgUrl,
+      imdbUrl: movie.imdbUrl,
+      imdbId: movie.imdbId,
+    };
```
ivbutsykin, line 20: Hm, can we remove `newMovie` and simple use `movie` from argument?

BTW you can make object in such way:
```
const newMovie = { ....movie };
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,89 @@
...
+        imdbUrl,
+        imdbId,
+      });
+
+      setCount(prevCount => prevCount + 1);
+      setTitle('');
+      setDescription('');
+      setImgUrl('');
+      setImdbUrl('');
+      setImdbId('');
```
ivbutsykin, line 36: I can recommend move it in separate `reset` function

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,89 @@
...
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const handleFormSubmit = (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (title && imgUrl && imdbUrl && imdbId) {
```
ivbutsykin, line 22: Also, to avoid nesting, you can write in such way:
```
if (!(title && imgUrl && imdbUrl && imdbId)) {
    return;
}

onAdd(...);
...
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,89 @@
...
+      setCount(prevCount => prevCount + 1);
+      setTitle('');
+      setDescription('');
+      setImgUrl('');
+      setImdbUrl('');
+      setImdbId('');
+    }
+  };
+
+  const disaibled = !title || !imgUrl || !imdbUrl || !imdbId;
```
ivbutsykin, line 40: Does this condition relevant to this?
```
title && imgUrl && imdbUrl && imdbId
```

Can we move it above and reuse for `handleFormSubmit` function?

ivbutsykin (changes requested):

src/App.tsx
```diff
@@ -1,16 +1,23 @@
...
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie: Movie) => {
+          setMovies([...movies, movie]);
+        }}
```
Viktor-Kost, line 19: it's better to move this handler to an own function

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+    event.preventDefault();
+    onAdd(movie);
+    setMovie({
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
+    setCount(count + 1);
```
Viktor-Kost, line 36: plus `1` to a prev value

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+  const handleSubmit = (event: { preventDefault: () => void; }) => {
+    event.preventDefault();
+    onAdd(movie);
+    setMovie({
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
```
Viktor-Kost, line 35: looks like a reset function. So move this part of code to reset function

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+    setCount(count + 1);
+  };
+
+  const isButtonDisabled = () => {
```
Viktor-Kost, line 39: no needs to create as a function, just make a simple variable

src/components/NewMovie/NewMovie.tsx
Viktor-Kost: isn't fixed

DarkMistyRoom (changes requested):
Good job! Fix only this moment: 
I fill all fields but add button is inactive and there is no message about what's wrong
image

Anastasiia-Svintsova (approved):

PR #1449, opened 2023-04-30
solution

AnastassiaH:
[DEMO LINK](https://anastassiah.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,53 +1,113 @@
...
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
+  const [newMovie, setNewMovie] = useState(emptyMovie);
+  const [isAddDisabled, setIsAddDisabled] = useState(true);
+
+  useEffect(() => {
+    setIsAddDisabled(() => {
+      return Object.values(newMovie).some((el, i) => (i === 1 ? false : !el));
```
TarasHoliuk, line 30: Instead of checking index 'i', you may directly check if the key is 'description'.

TarasHoliuk, line 30: Use a more descriptive variable name instead of 'el', like 'fieldValue'.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,53 +1,113 @@
...
+    e.preventDefault();
+    onAdd(newMovie);
+    setCount(prev => prev + 1);
+    setNewMovie(emptyMovie);
+  };
+
+  const validateCustom = ({ dataset, value }: HTMLInputElement) => {
+    const targetKey = dataset.cy?.split('-')[1];
+    // eslint-disable-next-line
+    const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
```
TarasHoliuk, line 70: Move the URL validation pattern to a constant outside the component definition.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,53 +1,113 @@
...
+
+    // write custom requirements
+    switch (targetKey) {
+      case MovieField.imdbUrl:
+      case MovieField.imgUrl:
+      case MovieField.imdbId:
```
TarasHoliuk, line 76: Remove the case for 'MovieField.imdbId' as it is not a URL.

src/components/TextField/TextField.tsx
```diff
@@ -19,13 +20,17 @@ export const TextField: React.FC<Props> = ({
   label = name,
   required = false,
   onChange = () => {},
+  validateCustom = () => {},
```
TarasHoliuk, line 23: The default value for validateCustom should be a function that always returns true. Change it to: validateCustom = () => true.

src/components/TextField/TextField.tsx
```diff
@@ -19,13 +20,17 @@ export const TextField: React.FC<Props> = ({
...
 }) => {
-  // generage a unique id once on component load
   const [id] = useState(() => `${name}-${getRandomDigits()}`);
-
-  // To show errors only if the field was touched (onBlur)
   const [touched, setToched] = useState(false);
-  const hasError = touched && required && !value;
+  const [customValid, setCustomValid] = useState(false);
```
TarasHoliuk, line 27: The initial state for customValid should be true instead of false, as the field is not invalid by default. Change it to: const [customValid, setCustomValid] = useState(true);

src/components/TextField/TextField.tsx
```diff
@@ -7,6 +7,7 @@ type Props = {
   label?: string,
   required?: boolean,
   onChange?: (newValue: string) => void,
+  validateCustom: (target: HTMLInputElement) => boolean,
```
TarasHoliuk, line 10: The validateCustom prop should be optional (you are setting the default value for it), add a question mark to make it optional: validateCustom?: (target: HTMLInputElement) => boolean.

src/components/TextField/TextField.tsx
```diff
@@ -19,13 +20,17 @@ export const TextField: React.FC<Props> = ({
...
 }) => {
-  // generage a unique id once on component load
   const [id] = useState(() => `${name}-${getRandomDigits()}`);
-
-  // To show errors only if the field was touched (onBlur)
   const [touched, setToched] = useState(false);
-  const hasError = touched && required && !value;
+  const [customValid, setCustomValid] = useState(false);
+
+  const hasError = (touched && required && !value) || customValid;
```
TarasHoliuk, line 29: The hasError condition should check if customValid is false, not true. Change the condition to: const hasError = (touched && required && !value) || !customValid;

TarasHoliuk (changes requested):

src/components/TextField/TextField.tsx
AnastassiaH: це випливає з одної з попередніх правок

src/components/TextField/TextField.tsx
```diff
@@ -43,13 +48,24 @@ export const TextField: React.FC<Props> = ({
           type="text"
           placeholder={`Enter ${label}`}
           value={value}
-          onChange={event => onChange(event.target.value)}
-          onBlur={() => setToched(true)}
+          onChange={event => {
+            onChange(event.target.value);
+            checkCustom(event.target);
+            setCustomValid(true);
+          }}
```
polosanya, line 55: Create a separate function to handle this change. Also, "onChange" name for function from props is confusing

polosanya (changes requested):
Make sure that you added all needed files to the commit
image

src/components/TextField/TextField.tsx
AnastassiaH: функції onChange були в початковому файлі

src/components/TextField/TextField.tsx
```diff
@@ -43,13 +54,20 @@ export const TextField: React.FC<Props> = ({
           type="text"
           placeholder={`Enter ${label}`}
           value={value}
-          onChange={event => onChange(event.target.value)}
-          onBlur={() => setToched(true)}
+          onChange={event => onTextFieldChange(event.target)}
```
VitaliyBondarenko1982, line 57:
```suggestion
          onChange={onTextFieldChange}
```

src/components/TextField/TextField.tsx
```diff
@@ -19,13 +20,23 @@ export const TextField: React.FC<Props> = ({
...
+
+  const hasError = (touched && required && !value) || !customValid;
+
+  const checkCustom = (target: HTMLInputElement) => {
+    setCustomValid(validateCustom(target));
+  };
+
+  const onTextFieldChange = (target: EventTarget & HTMLInputElement) => {
+    onChange(target.value);
+    checkCustom(target);
```
VitaliyBondarenko1982, line 37:
```suggestion
  const onTextFieldChange = (event: React.ChangeEvent<HTMLInputElement>) => {
    onChange(event.target.value);
    checkCustom(event.target);
```

src/components/TextField/TextField.tsx
```diff
@@ -43,13 +54,20 @@ export const TextField: React.FC<Props> = ({
           type="text"
           placeholder={`Enter ${label}`}
           value={value}
-          onChange={event => onChange(event.target.value)}
-          onBlur={() => setToched(true)}
+          onChange={event => onTextFieldChange(event.target)}
+          onBlur={(event) => {
```
VitaliyBondarenko1982, line 58: here also better to create handler and pass reference here

VitaliyBondarenko1982 (approved):
Good job.
Check my comments, please, about anonymous functions in jsx and refactoring onChange handler

PR #1469, opened 2023-05-02
Develop

vinnykalisa:
[DEMO LINK](https://vinnykalisa.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,95 @@
-import { useState } from 'react';
+import { FC, FormEvent, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+interface Props {
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,93 @@
...
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    };
+
+    onAdd(movie);
+
+    setTitle('');
+    setDescription('');
```
SerhiiKirik, line 38: would be better if you create a separate function to reset form

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,93 @@
...
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(newTitle) => setTitle(newTitle)}
```
SerhiiKirik, line 58: it's the same
In all cases
```suggestion
        onChange={setTitle}
```

SerhiiKirik (changes requested):
Well done 🔥 
Let's fix some bugs 🧑‍💻 
I can add movies just with spaces in the fields
image

src/components/NewMovie/NewMovie.tsx
SerhiiKirik: why you ignore it?

src/components/NewMovie/NewMovie.tsx
SerhiiKirik: do it in all cases

SerhiiKirik (approved):
Good job 🔥

PR #1476, opened 2023-05-04
Solution

emxm:
https://emxm.github.io/react_movies-list-add-form/


tetiana-ivanchenko (changes requested):
now I can add film with just spaces in all fileds

emxm:
Thanks for your review, seems to be working as it should

witflash (approved):
Well done 👍 
Could you also try to implement optional URL validation?
image

It'll be useful for your future work.

PR #1478, opened 2023-05-07
solution

gabriella135:
[DEMO LINK](https://gabriella135.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,93 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (NewMovie: Movie) => void
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const resetForm = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
+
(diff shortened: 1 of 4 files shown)


OleziO (changes requested):

polosanya (changes requested):
It is impossible to type any values in inputs
image
Also, some errors duplicate

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,93 @@
...
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [newMovie, setNewMovie] = useState({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
```
OleziO, line 17:
```
const initMovie = {
    title: '',
    description: '',
    imgUrl: '',
    imdbUrl: '',
    imdbId: '',
  }
  
const [newMovie, setNewMovie] = useState(initMovie);

const resetForm = () => {
    setNewMovie(initMovie);
  };
```

OleziO: Make default movie value to avoid code dublication

OleziO (changes requested):

polosanya (changes requested):
Why it's not possible to type ' '(space) in inputs? Fix this problem to pass the tests
image
image

sTorba24 (changes requested):
Good work! Lets fix bug you have

It gives an error but i still can add movie, use this validation to prevent adding
image

maxim2310 (changes requested):
Good job, but you need to fix all comments
1. this problem still exists 
image
2. and this
image

gabriella135:
I fixed the first problem, but I need to leave each element separately with useState, because if I make a common object, then I have problems with the site and the inputs don't work correctly.

maxim2310 (changes requested):
try to make new movie state like one obj its not so complicated, if you will have questions, you can ask in chat

anastasiia-tilikina (changes requested):
Almost done!

the validator should show one error at a time either field is required or invalid url. let's fix it.
image

here i still have errors but submit button is not disabled anymore(
image

maxim2310 (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)
there should be a require error firstly 
image
2.

PR #1482, opened 2023-05-09
solution

Oleh-Rybachuk:
https://oleh-rybachuk.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,103 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,103 @@
...
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    });
+
+    resetForm();
+    setCount(currentCount => currentCount + 1);
+  };
+
+  const checkValid = (str:string) => {
```
Viktor-Kost, line 40: it's better to name sth like this `checkIsUrlValid`
And I'd move this helper function outside the component since there is no direct dependence

src/components/TextField/TextField.tsx
```diff
@@ -18,14 +18,18 @@ export const TextField: React.FC<Props> = ({
...
-  onChange = () => {},
+  onChange,
 }) => {
   // generage a unique id once on component load
   const [id] = useState(() => `${name}-${getRandomDigits()}`);
 
   // To show errors only if the field was touched (onBlur)
-  const [touched, setToched] = useState(false);
-  const hasError = touched && required && !value;
+  const [touched, setTouched] = useState(false);
```
Viktor-Kost, line 27:
```suggestion
  const [isTouched, setIsTouched] = useState(false);
```

Viktor-Kost (approved):
Well done! 👏

PR #1484, opened 2023-05-10
Solution

andrii-kovalskyi02:
[DEMO LINK](https://andrii-kovalskyi02.github.io/react_movies-list-add-form/)


tetiana-ivanchenko (changes requested):
you have to check the url after the user clicks the add button because it's not obvious why the button is disabled

anstsot (approved):
Great job, approved

PR #1487, opened 2023-05-11
Develop

LBondarchuk:
[DEMO LINK](https://LBondarchuk.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,19 @@
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
 
 export const App = () => {
+  const [movies, setMovies] = useState([...moviesFromServer]);
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={setMovies} movies={movies} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
 import { useState } from 'react';
 import { TextField } from '../TextField';
 import { Movie } from '../../types/Movie';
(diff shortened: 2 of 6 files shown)


src/App.tsx
```diff
@@ -1,16 +1,19 @@
...
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={setMovies} movies={movies} />
```
polosanya, line 16: NewMovie component should not directly update the movies state. Instead, create a new function that adds a movie to the list and updates the state.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,53 +1,78 @@
...
-        name="imdbId"
-        label="Imdb ID"
-        value=""
+        key={inputName}
+        name={inputName}
+        label={inputName}
+        value={movie[inputName]}
+        required={inputName !== 'description'}
+        isEqual={['imgUrl', 'imdbUrl'].includes(inputName) ? validUrl(movie[inputName]) : true}
+        // eslint-disable-next-line react/jsx-no-bind
```
polosanya, line 55: Avoid using anonymous functions in props. You can define a separate function for handling the onChange event and pass it to the TextField component. This way, you can prevent unnecessary re-renders.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,53 +1,78 @@
...
+    setCount(movies.length);
+  };
 
-      <TextField
-        name="imdbUrl"
-        label="Imdb URL"
-        value=""
-      />
+  function onChange(name: string, value: string) {
+    setMovie({ ...movie, ...{ [name]: value } });
```
polosanya, line 33: Instead of spreading the movie object and an additional object with the updated field, you can directly update the movie object with the new name-value pair. Replace `setMovie({ ...movie, ...{ [name]: value } });` with `setMovie({ ...movie, [name]: value });`.

src/components/TextField/TextField.tsx
```diff
@@ -42,15 +47,16 @@ export const TextField: React.FC<Props> = ({
           })}
           type="text"
           placeholder={`Enter ${label}`}
-          value={value}
-          onChange={event => onChange(event.target.value)}
+          defaultValue={value}
+          onChange={event => onChange(name, event.currentTarget.value.trim())}
```
polosanya, line 51: What's the purpose of using currentTarget instead of target?

polosanya (changes requested):
Good job! Let's improve your code a bit.
Also, pay attention that all text should be capitalized
image
image

src/components/TextField/TextField.tsx
LBondarchuk: Уarlier I received information that it is more correct. Maybe not in this case. Maybe I got the wrong information

tetiana-ivanchenko (approved):

PR #1489, opened 2023-05-12
Solution

Zibi95:
[DEMO LINK](https://Zibi95.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,93 @@
-import { useState } from 'react';
+/* eslint-disable implicit-arrow-linebreak */
+import { ChangeEvent, FormEventHandler, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type NewMovieProps = {
+  onAdd: (movie: Movie) => void;
+};
+
+const initialFormValues: Movie = {
+  title: '',
+  description: '',
+  imgUrl: '',
+  imdbUrl: '',
+  imdbId: '',
+};
+
+export const NewMovie = ({ onAdd }: NewMovieProps) => {
+  const [formValues, setFormValues] = useState(initialFormValues);
+  const [count, setCount] = useState(0);
+
+  const {
+    title, description, imdbId, imdbUrl, imgUrl,
+  } = formValues;
+
+  const onChange = ({ target }: ChangeEvent<HTMLInputElement>) => {
+    const { name, value } = target;
+
+    setFormValues(prevState => ({
+      ...prevState,
+      [name]: value,
+    }));
+  };
+
+  const onSubmit: FormEventHandler = (e) => {
+    e.preventDefault();
+
+    onAdd(formValues);
+
+    setFormValues(initialFormValues);
+    setCount(prevState => prevState + 1);
+  };
+
+  const isDisabled = !title || !imdbId || !imdbUrl || !imgUrl;
 
   return (
-    <form className="NewMovie" key={count}>
+    <form
+      className="NewMovie"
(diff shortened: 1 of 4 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -20,12 +26,12 @@ export const TextField: React.FC<Props> = ({
   required = false,
   onChange = () => {},
 }) => {
-  // generage a unique id once on component load
   const [id] = useState(() => `${name}-${getRandomDigits()}`);
```
Radoslaw-Czerniawski, line 29: It would be better to generate the id with [useMemo](https://react.dev/reference/react/useMemo) rather than this monstrosity, if this is a construct made by mate academy feel free to read about how to use useMemo and implement it instead of this

src/components/TextField/TextField.tsx
```diff
@@ -1,17 +1,23 @@
...
+
+export const checkIfUrl = (url: string) => {
+  const pattern
+    // eslint-disable-next-line max-len
+    = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
+
+  return pattern.test(url);
+};
 
 export const TextField: React.FC<Props> = ({
```
Radoslaw-Czerniawski, line 22: Same case as before with an InputField, according to S.O.L.I.D. principles (Liskov Substitution to be exact) it should be possible to switch around this supertype object class with the subtype object class (this component and input component). What I mean by that is you should be able to pass the same props to this component as to the input, this can be done the same way as you did in some examples before with spreading the `...rest` of the props on the input, only explicitly providing attributes that you modify and extending the interface with input type props.
```tsx
interface TextFieldProps extends HTMLProps<HTMLInputElement> { };
```
or if you want to use type

```tsx
type TextFieldProps  = {} & HTMLProps<HTMLInputElement>;
```

You could actually just use same props for this component as are used on the input 
```tsx
export const TextField: React.FC<HTMLProps<HTMLInputElement>>
```
[Here's a video](https://www.youtube.com/watch?v=MSq_DCRxOxw) on the topic of SOLID principles in React

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,93 @@
...
+    || !imgUrl
+    || !checkIfUrl(imdbUrl)
+    || !checkIfUrl(imdbUrl);
 
   return (
-    <form className="NewMovie" key={count}>
+    <form className="NewMovie" key={count} onSubmit={onSubmit}>
       <h2 className="title">Add a movie</h2>
 
       <TextField
```
Radoslaw-Czerniawski, line 54: You could think on some way to map those Textfields rather than declare them one by one, maybe based on the keys of initialFormValues?

Radoslaw-Czerniawski (changes requested):
Generally good pr, left some suggestions, if you don't want to tinker with them and go further apply for re-review I will accept it

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,53 +1,85 @@
...
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const isDisabled = !title
+    || !imdbId
+    || !imdbUrl
+    || !imgUrl
+    || !checkIfUrl(imdbUrl)
```
darokrk, line 50: You have repeated the same condition twice: `!checkIfUrl(imdbUrl)`. One of them should be `!checkIfUrl(imgUrl)`.

darokrk (approved):
Looks nice, good job 🎖️ ! You used one condition twice, so look to fix it and can move forward :) 👍🏽

PR #1495, opened 2023-05-14
solution

Nikoramo:
[DEMO LINK](https://Nikoramo.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,91 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (newMovie: Movie) => void
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,91 @@
...
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const resetForm = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
```
ivbutsykin, line 23: Nice decision to move it in separate function!

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,91 @@
...
+      imdbId,
+    };
+
+    onAdd(newMovie);
+
+    resetForm();
+    setCount(prevCount => prevCount + 1);
+  };
+
+  const submitDisabled = !title.trim() || !imgUrl.trim()
```
ivbutsykin, line 41: I can recommend add `is` suffix for `boolean` types:
```suggestion
  const isSubmitDisabled = !title.trim() || !imgUrl.trim()
```

ivbutsykin (approved):
Well done!

You can implement one improvement related on optional validation task: if url has invalid format then add button should be disabled.

PR #1496, opened 2023-05-15
solution

MikiOOOO:
https://MikiOOOO.github.io/react_movies-list-add-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,27 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [currentMoviesList,
+    setCurrentMoviesList] = useState([...moviesFromServer]);
+
+  const addMovie = (movie: Movie) => {
+    setCurrentMoviesList(previousMoviesList => {
+      return [...previousMoviesList, movie];
+    });
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={currentMoviesList} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addMovie} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,125 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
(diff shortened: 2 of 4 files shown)


src/App.tsx
```diff
@@ -1,16 +1,27 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [currentMoviesList,
+    setCurrentMoviesList] = useState([...moviesFromServer]);
```
d-wasilewski, line 10: You don't have to copy it, you use it only as initial value, so you won't have any problem with referencing it later
```suggestion
    setCurrentMoviesList] = useState(moviesFromServer);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,125 @@
...
+  const defaultMovie = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
+
+  const [count, setCount] = useState(0);
+  const [movie, setMovie] = useState<Movie>({ ...defaultMovie });
```
d-wasilewski, line 21: again, you don't have to spread it

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,125 @@
...
+
+  const {
+    title,
+    description,
+    imgUrl,
+    imdbId,
+    imdbUrl,
+  } = movie;
+
+  const isValid = title.length && imgUrl.length
```
d-wasilewski, line 31: Empty string is always falsy itself
```suggestion
  const isValid = title&& imgUrl
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,125 @@
...
+      && imdbUrl.length && imdbId.length;
+
+  const submitMovie = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    // const isValid = title.length && imgUrl.length
+    //   && imdbUrl.length && imdbId.length;
```
d-wasilewski, line 38: remove unnecessaryh comments

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,125 @@
...
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const defaultMovie = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
```
d-wasilewski, line 18: Both approaches have their uses, but it's generally better to keep them in separate useStates instead of a single object

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1502, opened 2023-05-20
Task12

Ryshon:
[DEMO LINK](https://Ryshon.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,27 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,16 +1,27 @@
...
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const onAdd = (newMovie: Movie) => {
```
polosanya, line 11:
```suggestion
  const addMovie = (newMovie: Movie) => {
```

polosanya (changes requested):
There should be errors on blur if input is empty. Check the working example
image
image

PR #1507, opened 2023-05-23
add solution

PufAnimalsDev:
[DEMO LINK](https://pufanimalsdev.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,105 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+// eslint-disable-next-line max-len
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,105 @@
...
+
+export const NewMovie: React.FC<NewMovieProps> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [newMovie, setNewMovie] = useState({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
```
d-wasilewski, line 20: Both approaches have their uses, but it's generally better to keep them in separate useStates

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,105 @@
...
+
+  const validateUrl = (url: string) => {
+    return (url.match(pattern) || [])[0] === url;
```
d-wasilewski, line 23: use `.test` method

d-wasilewski (approved):

PR #1512, opened 2023-05-23
Solution

nazarkharkevych:
https://nazarkharkevych.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -24,16 +24,11 @@ export const TextField: React.FC<Props> = ({
 }) => {
   const [id] = useState(() => `${name}-${getRandomDigits()}`);
   const [touched, setToched] = useState(false);
-  const [isValidUrl, setIsValidUrl] = useState(true);
   const hasError = touched && required && !value.trim();
-  const notValidUrl = touched && value.trim() && !isValidUrl;
+  const notValidUrl = touched && value.trim() && !onValidateUrl(value);
 
   const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     onChange(e.target.value);
-
-    const isValid = onValidateUrl(e.target.value);
-
-    setIsValidUrl(isValid);
   };
 
   return (


src/App.tsx
```diff
@@ -1,16 +1,27 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const handleOnAdd = (movie: Movie) => {
```
loralevitska, line 11: https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364
```suggestion
  const handleAdd = (movie: Movie) => {
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,110 @@
...
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(newTitle) => setTitle(newTitle)}
```
loralevitska, line 71: if the parameter does not need to be processed, it will still get to the function as a parameter. Fix all cases
```suggestion
        onChange={setTitle}
```

loralevitska (changes requested):
Keep up the good work! Some comments to fix:
I can add movie with empty title. Fix it with [String.prototype.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim)
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,110 @@
...
+      imgUrl: imageUrl,
+      imdbUrl,
+      imdbId,
+    };
+
+    onAdd(newMovie);
+
+    resetForm();
+
+    setCount(count + 1);
```
DarkMistyRoom, line 56: isn't it work without count incrementing? `resetForm` refreshes the state and should trigger rendering

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,110 @@
...
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imageUrl, setImageUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [imgUrlIsValid, setImgUrlIsValid] = useState(true);
+  const [imdbUrlIsValid, setImdbUrlIsValid] = useState(true);
```
DarkMistyRoom, line 19: could we move these values out of the state?

DarkMistyRoom (changes requested):
Good job! But looks like the `state` can be reduced

SanyaBratashchuk (approved):

PR #1520, opened 2023-05-25
Solution

NadieinOleh:
https://NadieinOleh.github.io/react_movies-list-add-form/


anstsot (changes requested):
Good work but it's possible to add empty movie card. Please add validation
image

PR #1521, opened 2023-05-25
Movie list form (react app) - solution

ctgdprzyb:
[DEMO LINK](https://ctgdprzyb.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,10 +1,15 @@
-import { useState } from 'react';
+import { FormEvent, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+interface NewMovieProps {
+  onAdd: (arg0: Movie) => void,
+}
+
+export const NewMovie = ({ onAdd }: NewMovieProps) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
         required
       />
 
       <TextField
         name="description"
         label="Description"
-        value=""
+        value={description}
+        onChange={(input) => {
+          setDescription(input);
```
d-wasilewski, line 61: Try this again, maybe this time it will work 🙇 
```suggestion
        onChange={setDescription};
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+
+  const movie: Movie = {
+    title,
+    description,
+    imgUrl: imageUrl,
+    imdbUrl,
+    imdbId,
+  };
+
+  const handleSubmit = (event: FormEvent<HTMLFormElement>, newMovie: Movie) => {
```
d-wasilewski, line 30: You can move this movie object inside this function. This way it will be created only on the function invocation
```suggestion
  const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
```

d-wasilewski (changes requested):
Form should be cleared after the movie has been submited

d-wasilewski (approved):

PR #1529, opened 2023-06-06
add movies form

DidenkoDenis306:
[DEMO LINK](https://DidenkoDenis306.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -33,18 +33,11 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
     setCount(current => current + 1);
   };
 
-  const newMovie = {
-    title,
-    description,
-    imgUrl,
-    imdbUrl,
-    imdbId,
-  };
-
   const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
     event.preventDefault();
 
-    onAdd(newMovie);
+    onAdd(formState);
+    
     clearForm();
   };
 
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,93 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
33Orange, line 17: You can use only 1 state for all fields.
Create const outside your component where you can save your `initialFormState` which will have all initial values 

```
const initialFormState = {
  title: '',
  ...
};
```

```suggestion
  const [count, setCount] = useState(0);
  const [formState, setFormState] = useState(initialFromState);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,93 @@
...
+  const enteredRequiredFields = title && imgUrl && imdbUrl && imdbId;
+
+  const clearForm = () => {
+    setCount(count + 1);
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
```
33Orange, line 28: If you add `initialFormState` and one state for all fields as I said - you can set initial state here and that's all

33Orange (changes requested):
Almost done

A few moments to improve

33Orange (approved):
Well done

PR #1534, opened 2023-06-07
add task solution

maksym-nemera:
[DEMO LINK](https://maksym-nemera.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,7 +8,7 @@ import { Movie } from './types/Movie';
 export const App = () => {
   const [movies, setMovies] = useState(moviesFromServer);
 
-  const handleAddMovie = (movie: Movie) => {
+  const addMovie = (movie: Movie) => {
     setMovies([...movies, movie]);
   };
 
@@ -18,7 +18,7 @@ export const App = () => {
         <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie onAdd={handleAddMovie} />
+        <NewMovie onAdd={addMovie} />
       </div>
     </div>
   );
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const handleAddMovie = (movie: Movie) => {
+    setMovies([...movies, movie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const handleAddMovie = (movie: Movie) => {
```
ruslanpashkov, line 11: it's not a handler, so you don't need `handle` keyword here
```suggestion
  const addMovie = (movie: Movie) => {
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,118 @@
-import { useState } from 'react';
+import { FC, useEffect, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void,
```
ruslanpashkov, line 6:
```suggestion
  onAdd: (movie: Movie) => void;
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,118 @@
...
+      setDisabledButton(false);
+    }
+  }, [title, imgUrl, imdbUrl, imdbId]);
+
+  const resetForm = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
```
ruslanpashkov, line 40:
```suggestion
    setFormData(initialFormData);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,118 @@
...
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    setCount(count + 1);
```
ruslanpashkov, line 46: we change our current state based on previous state, let's use functional `setCount` here

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,118 @@
...
+    if (count > 0) {
+      onAdd(formData as Movie);
+    }
+  }, [count]);
+
+  useEffect(() => {
+    if (title && imgUrl && imdbUrl && imdbId) {
+      setDisabledButton(false);
+    }
+  }, [title, imgUrl, imdbUrl, imdbId]);
```
ruslanpashkov, line 33: what if the fields are only filled with spaces?

also we can create separate variable for it and pass negation of this value to `disabled` attribute

ruslanpashkov (changes requested):

silchyn (changes requested):
Form can be submitted several times in a row with no values entered

silchyn (changes requested):
When all values are spaces, submit button is disabled, but no validation message is shown
image

And678 (approved):
Good work, approve :heavy_check_mark:

PR #1541, opened 2023-06-16
MoviesForm

IvanFront-end:
[DEMO LINK](https://IvanFront-end.github.io/react_movies-list-add-form/)

Can you show please, how to add patter into function? Want to know, for myself.... because this solution was dificult for me.... I did this with mentor, he helped me..


sTorba24 (approved):
Good job! 
if you mean how to add pattern so i explain a little bit

create pattern - you already have it in the task description
To check if smth match pattern you can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test
create function that will get string and test it

call it for your inputs

PR #1551, opened 2023-06-21
add movies list add form solution

opiquem:
[DEMO LINK](https://maksym-bobryk.github.io/react_movies-list-add-form/)


33Orange (changes requested):
Almost done

I can push new movie with spaces in fields
image

dinaminahi (approved):
Well done 👍

PR #1564, opened 2023-06-22
solution

maryanmats:
[DEMO LINK](https://maryanmats.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,34 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [visibleMovies, setVisibleMovies] = useState(moviesFromServer);
+
+  const addMovieHandler = ({
+    title, description, imgUrl, imdbUrl, imdbId,
+  }: Movie) => {
+    const movieToAdd = {
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    };
+
+    setVisibleMovies(prevMovies => [...prevMovies, movieToAdd]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={visibleMovies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addMovieHandler} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,85 @@
-import { useState } from 'react';
+import { FC, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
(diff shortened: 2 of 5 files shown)


src/App.tsx
```diff
@@ -1,16 +1,34 @@
...
+    const movieToAdd = {
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    };
+
+    setVisibleMovies(prevMovies => [...prevMovies, movieToAdd]);
+  };
```
dinaminahi, line 23:
```suggestion
  const addMovie = (movie: Movie) => {
    setVisibleMovies(prevMovies => [...prevMovies, movie]);
  };
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,85 @@
...
+export const NewMovie: FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
dinaminahi, line 17: Instead of creating so many states, let's create one that will be object `formState` and will contain all fields as properties:
`const [formState, setFormState] = useState({ title: "", description: "", imgUrl: "", ... });`

dinaminahi (changes requested):
Good job 👍 
Few fixes needed

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,94 @@
...
+};
+
+export const NewMovie: FC<Props> = (props: Props) => {
+  const [count, setCount] = useState(0);
+
+  const [formState, setFormState] = useState(defaultFormState);
+
+  const { onAdd } = props;
+
+  const isSubmitButtonDisabled = (
```
roman-mirzoian, line 24: The logic of checking the validity of all fields can be transferred to a separate helper function outside the component (`checkFormFields()`), pass `formState` there and immediately return the result of the check.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,94 @@
...
+      onSubmit={handleSubmit}
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={formState.title}
```
roman-mirzoian, line 58: It would be cool to use destructuring to use such fields:
```
const { title, description, etc. } = formState;
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,94 @@
-import { useState } from 'react';
+import { FC, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
```
roman-mirzoian, line 5: Similar names are okay for a small project, but on large projects we will have many interfaces and types, so it is better to give more detailed names, for example `NewMovieProps` or `NewMovieType`.
This applies to all code.

roman-mirzoian (changes requested):
A good job has been done and only a few minor fixes remain to be made.

naziamoff (approved):
👍

PR #1567, opened 2023-06-22
solution

kirilshorin:
[DEMO LINK](https://kirilshorin.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,95 @@
-import { useState } from 'react';
+import { FC, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isValid = title.trim()
+    && imgUrl.trim()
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+}
+
+export const NewMovie: FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
33Orange, line 16: You can use only 1 state for all fields.
Create const outside your component where you can save your `initialFormState` which will have all initial values 

```
const initialFormState = {
  title: '',
  ...
};
```

And find the way how to handle your changes to formState

```suggestion
  const [formState, setFormState] = useState(initialFromState);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+    && imgUrl.trim()
+    && imdbUrl.trim()
+    && imdbId.trim();
+
+  const clearState = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
```
33Orange, line 28: You can set initial state here and that's all

```suggestion
     setFormState(initialFormState)
```

33Orange (changes requested):
Almost done

A few moments to improve

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,106 @@
...
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const newMovie = {
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    };
```
33Orange, line 51: Do you need this object? You have same object in `formState` state

33Orange (changes requested):
Just a one moment

dffUqp (approved):
LGTM:fire:

PR #1569, opened 2023-06-22
add solution

growingananas:
[DEMO LINK](https://growingananas.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,102 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (movie: Movie) => void
+}
+
+const defaultFormValues = {
+  title: '',
+  description: '',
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,102 @@
...
+    imdbUrl,
+    imdbId,
+  } = formValues;
+
+  const clearFormAfterSubmit = () => {
+    setFormValues(defaultFormValues);
+    setCount((prev) => prev + 1);
+  };
+
+  const isSubmitButtonDisabled = !title && !imgUrl && !imdbUrl && !imdbId;
```
33Orange, line 34: Your button activates when I add spaces to any field but other fields still empty.
I can push new movie with spaces in fields

image
image

33Orange (changes requested):
Almost done

src/components/NewMovie/NewMovie.tsx
growingananas: got it!

growingananas: btw, on demo version, one can also add a movie with all empty fields 
image

silchyn (changes requested):
The form shouldn't be submittable when all values are spaces

growingananas:
> The form shouldn't be submittable when all values are spaces

It must work now. Have a look, pls

dffUqp (approved):
LGTM:fire:

PR #1572, opened 2023-06-23
Develop

MaksimKad:
[DEMO LINK](https://MaksimKad.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,4 +1,4 @@
-import { React, useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
 import { Movie } from '../../types/Movie';
 
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,58 +1,127 @@
-import { useState } from 'react';
+import { React, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,58 +1,127 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+
+  const preparedForm: Movie = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
```
33Orange, line 18: Move this const ourside your component to avoid re-creating every render.
Use more descriptive name. `initialFormState` for example

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,58 +1,127 @@
...
+
+  const {
+    title,
+    description,
+    imgUrl,
+    imdbUrl,
+    imdbId,
+  } = movieForm;
+
+  const onChange = (key: string, value: string) => {
```
33Orange, line 30: onChange what?

```suggestion
  const handleFormChange  = (key: string, value: string) => {
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,58 +1,127 @@
...
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(value) => {
+          onChange('title', value);
+        }}
```
33Orange, line 69:
```suggestion
        onChange={value => onChange('title', value)}
```

33Orange (changes requested):
Almost done

33Orange (approved):
Well done

PR #1573, opened 2023-06-23
solution

skinnyjazz:
[DEMO LINK](https://skinnyjazz.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,27 @@
+import { FC, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
-export const App = () => {
+export const App: FC = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  const onAdd = (movie: Movie) => {
+    setMovies((prew) => [
+      ...prew,
+      movie,
+    ]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={onAdd} />
       </div>
     </div>
   );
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,16 +1,27 @@
...
+import { Movie } from './types/Movie';
+
+export const App: FC = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  const onAdd = (movie: Movie) => {
+    setMovies((prew) => [
+      ...prew,
+      movie,
+    ]);
```
33Orange, line 15: Use more descriptive names
```suggestion
    setMovies(prevMovies => [...prevMovies, movie ]);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,100 @@
...
+  title: '',
+  description: '',
+  imgUrl: '',
+  imdbUrl: '',
+  imdbId: '',
+};
+
+export const NewMovie: FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
```
33Orange, line 19: Remove redundant comments

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,100 @@
...
+
+  const {
+    title,
+    description,
+    imgUrl,
+    imdbUrl,
+    imdbId,
+  } = formState;
+
+  const chackedRequiredFields = title && imgUrl && imdbUrl && imdbId;
```
33Orange, line 31: I can push new movie with spaces in fields
image

33Orange (changes requested):
Almost done

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,106 @@
...
+
+  const chackedRequiredFields = (...requiredFields: string[]): boolean => {
+    let chackResult = true;
+
+    [...requiredFields].forEach(field => {
+      chackResult = chackResult && Boolean(field.trim());
+    });
+
+    return chackResult;
+  };
```
dffUqp, line 37: not bad solution, but let's replace this function with a const like this:
```jsx
  const isFieldsValid
    = title.trim() && imgUrl.trim() && imdbUrl.trim() && imdbId.trim();
```

dffUqp (approved):
LGTM:fire:

src/components/NewMovie/NewMovie.tsx
skinnyjazz: Why? I specifically wrote this function so as not to write such things.

dffUqp: this const is more optimized than your solution. 

now if you change some value you will call a forEach and then make some check, using this const you skip the forEach call and go straight to check.

PR #1575, opened 2023-06-23
add solution

YeMyr:
[DEMO LINK](https://yemyr.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/Helpers.ts
+++ b/src/Helpers.ts
@@ -1,0 +1,13 @@
+export function isValidHttpUrl(url: string) {
+  try {
+    const newUrl = new URL(url);
+
+    return newUrl.protocol === 'http:' || newUrl.protocol === 'https:';
+  } catch (err) {
+    return false;
+  }
+}
+
+// as it is optional, I've decided to use another solution
+// got it from here https://www.freecodecamp.org/news/how-to-validate-urls-in-javascript/
+// as for me, it's more convenient and readable
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,113 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
+import { isValidHttpUrl } from '../../Helpers';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImbdUrl] = useState('');
+  const [imdbId, setImbdId] = useState('');
+
+  const isRequired = title
+    && isValidHttpUrl(imgUrl)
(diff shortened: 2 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,113 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImbdUrl] = useState('');
+  const [imdbId, setImbdId] = useState('');
```
dffUqp, line 16: Can be replaced with one useState, using an object like in the example below.
```javascript
const [formData, setFormData] = useState({
  title: "",
  description: "",
  imgUrl: "",
  ...
});
```
Create const outside your component where you can save your `initialFormData` which will have all initial values
```jsx
const initialFormData = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: '',
};
```


To change this `formData` you can create a handler like this
```jsx
 const handleDataChange = (key: string, value: string) => {
    setFormData((currentState) => ({
      ...currentState,
      [key]: value,
    }));
  };
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,113 @@
...
+    setImbdUrl('');
+  };
+
+  const newMovie = {
+    title,
+    description,
+    imgUrl,
+    imdbUrl,
+    imdbId,
+  };
```
dffUqp, line 38: we update this variable on every render, put it in `handleSubmit`. 
(if we start using the `formData` object,  we can avoid to create this varriable)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,113 @@
...
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(value) => handleTitle(value)}
```
dffUqp, line 78: Just keep in mind that a components like `TextField` calls your function and provides all the props, so we don't need to create an extra function.
```suggestion 
        onChange={handleTitle}
```

src/Helpers.ts
```diff
@@ -0,0 +1,13 @@
...
+
+    return newUrl.protocol === 'http:' || newUrl.protocol === 'https:';
+  } catch (err) {
+    return false;
+  }
+}
+
+// as it is optional, I've decided to use another solution
+// got it from here https://www.freecodecamp.org/news/how-to-validate-urls-in-javascript/
+// as for me, it's more convenient and readable
```
dffUqp, line 13: good decision👍

dffUqp (changes requested):
LGTM:fire: but needs a few fixes:upside_down_face:


I can submit a form with spaces
image

src/components/NewMovie/NewMovie.tsx
YeMyr: fixed 👌

src/components/NewMovie/NewMovie.tsx
YeMyr: fixed 👌

src/components/NewMovie/NewMovie.tsx
YeMyr: fixed 👌

33Orange (approved):
Well done

PR #1576, opened 2023-06-23
add task solution

Yuliia-kucheriavenko:
[DEMO LINK](https://Yuliia-kucheriavenko.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,98 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+const initialFormState = {
+  title: '',
+  description: '',
+  imgUrl: '',
+  imdbUrl: '',
+  imdbId: '',
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [formState, setFormState] = useState(initialFormState);
+  const {
+    title,
+    description,
+    imgUrl,
+    imdbUrl,
+    imdbId,
+  } = formState;
+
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,98 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
```
roman-mirzoian, line 5: Similar names are okay for a small project, but on large projects we will have many interfaces and types, so it is better to give more detailed names, for example `NewMovieProps`.
It would also be really cool to keep similar types in one place, so you could move this type to the `types/Movie.ts` file and use only ready-made imports here. This would shorten the code a bit and make it cleaner.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,98 @@
...
+  const [formState, setFormState] = useState(initialFormState);
+  const {
+    title,
+    description,
+    imgUrl,
+    imdbUrl,
+    imdbId,
+  } = formState;
+
+  const isSubmitButtonActive = title && imgUrl && imdbUrl && imdbId;
```
roman-mirzoian, line 28: It is better to store `isActive` checks as a separate `state` so that there are no problems in use. The check itself is better to be done when updating `the state of the form` by simply going through the filled fields of the form and seeing if they exist (except for the `description`).

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,98 @@
...
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    onAdd(formState);
+
+    clearForm();
+  };
+
+  const changeFormState = (key: string, value: string) => {
+    setFormState((currentState) => ({
```
roman-mirzoian, line 44: Now if you put a space in all inputs, you can add a new card with all empty fields. Therefore, it is worth checking whether `value` is not an empty string or `trim` spaces manually.

roman-mirzoian (changes requested):
The work is done well and the component of the new movie itself remains a bit to be done.

PR #1584, opened 2023-06-26
add movies-list-add-form solution

vchervinko:
[DEMO LINK](https://vchervinko.github.io/react_movies-list-add-form/)


dffUqp (changes requested):
code looks very good:fire: but something is wrong with demo link

image

PR #1585, opened 2023-06-26
solution

thienlaoo:
DEMO LINK](https://thienlaoo.github.io/react_movies-list-add-form/)


dffUqp (changes requested):
Your demo link is not working but i see that code looks good only one moment to fix.

you don't handle a spaces so from can be submitted consisting of only spaces

PR #1590, opened 2023-06-30
react_movies-list-add-fom

vintwp:
https://vintwp.github.io/react_movies-list-add-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,15 +9,6 @@ import moviesFromServer from './api/movies.json';
 
 import './App.scss';
 
-/*
-
-Original title: The Umbrella Academy
-A family of former child heroes, now grown apart, must reunite to continue to protect the world.
-https://www.imdb.com/title/tt1312171/
-tt1312171
-https://m.media-amazon.com/images/M/MV5BNTZlNTY4ZGMtMzJjZC00NWFkLWFkZjItZDc2Y2Y1NGUyNzFhXkEyXkFqcGdeQXVyMTE5MTg5NDIw._V1_FMjpg_UY3000_.jpg
-*/
-
 export const App = () => {
   const initialMovies = [...moviesFromServer];
   const [moviesToRender, setMoviesToRender] = useState([...initialMovies]);
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,37 @@
-import './App.scss';
+import { useState } from 'react';
+
+import { Movie } from './types/Movie';
+
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+
 import moviesFromServer from './api/movies.json';
 
+import './App.scss';
+
+/*
+
+Original title: The Umbrella Academy
+A family of former child heroes, now grown apart, must reunite to continue to protect the world.
+https://www.imdb.com/title/tt1312171/
+tt1312171
+https://m.media-amazon.com/images/M/MV5BNTZlNTY4ZGMtMzJjZC00NWFkLWFkZjItZDc2Y2Y1NGUyNzFhXkEyXkFqcGdeQXVyMTE5MTg5NDIw._V1_FMjpg_UY3000_.jpg
+*/
+
 export const App = () => {
+  const initialMovies = [...moviesFromServer];
+  const [moviesToRender, setMoviesToRender] = useState([...initialMovies]);
+  const addMovie = (movie: Movie) => {
+    setMoviesToRender(currentMovies => [...currentMovies, movie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={moviesToRender} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addMovie} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,110 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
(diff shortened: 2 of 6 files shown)


src/App.tsx
```diff
@@ -1,16 +1,28 @@
...
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+
 import moviesFromServer from './api/movies.json';
 
+import './App.scss';
+
 export const App = () => {
+  const initialMovies = [...moviesFromServer];
+  const [moviesToRender, setMoviesToRender] = useState([...initialMovies]);
```
Viktor-Kost, line 14: As advised earlier, there is no need to create a copy of the moviesFromServer list. You can directly use the list in useState.

src/App.tsx
```diff
@@ -1,16 +1,28 @@
...
+
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+
 import moviesFromServer from './api/movies.json';
 
+import './App.scss';
+
 export const App = () => {
+  const initialMovies = [...moviesFromServer];
```
Viktor-Kost, line 13: Instead of creating a copy of the moviesFromServer list using the spread operator, you can directly use the list in useState.

src/App.tsx
```diff
@@ -1,16 +1,28 @@
...
+  const [moviesToRender, setMoviesToRender] = useState([...initialMovies]);
+  const addMovie = (movie: Movie) => {
+    setMoviesToRender(currentMovies => [...currentMovies, movie]);
```
Viktor-Kost, line 16: You should consider adding validation to ensure that the movie is not already in the moviesToRender list to avoid duplicate entries. You can check by IMDB ID

src/components/TextField/TextField.tsx
```diff
@@ -26,10 +29,28 @@ export const TextField: React.FC<Props> = ({
...
+
+      return;
+    }
+
+    switch (true) {
+      case !fieldData:
+        setFieldErrorMessage(`${label} is required`);
+        break;
+      case (fieldData.length < 3) && !isUrl:
+        setFieldErrorMessage(`${label} requires min 2 characters`);
```
Viktor-Kost, line 46: The error message states that the field requires min 2 characters, but the code checks that the `fieldData.length` is less than 3. This is a bit misleading and could lead to confusion. Update the condition or the error message to match each other.

src/components/TextField/TextField.tsx
```diff
@@ -26,10 +29,28 @@ export const TextField: React.FC<Props> = ({
 }) => {
   // generage a unique id once on component load
   const [id] = useState(() => `${name}-${getRandomDigits()}`);
+  const [fieldErrorMessage, setFieldErrorMessage] = useState('');
+
+  const validateTextField = (fieldData: string, isUrl = false): void => {
+    if (isUrl && !fieldData.match(pattern) && fieldData) {
```
Viktor-Kost, line 35: The condition in the if statement is not optimal. It should first check if `fieldData` is not empty before trying to match the pattern. This way, you avoid unnecessary regex matching that can be costly in terms of performance.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,100 @@
...
+        handlerOnSubmit(event);
+      }}
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
```
Viktor-Kost, line 17: You are missing the form validation as per requirements. You should also implement regex pattern validation for the image URL and IMDB URL fields.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,100 @@
...
+  const clearFormFields = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
+
+  const handlerOnSubmit = (event:React.FormEvent) => {
+    event.preventDefault();
```
Viktor-Kost, line 28: You are clearing the form fields before adding the movie. This does not cause a bug, but it's better to clear the fields after the movie has been added successfully.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +103,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!title || !imgUrl || !imdbUrl || !imdbId}
```
Viktor-Kost, line 106: You have disabled the submit button if any of the required fields are empty. This is good, but the submit button should also be disabled during form submission to prevent multiple submissions.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,100 @@
...
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
+
+  const handlerOnSubmit = (event:React.FormEvent) => {
+    event.preventDefault();
+    clearFormFields();
```
Viktor-Kost, line 29: The count state variable is incremented after clearing the form. This could lead to unnecessary re-renders if the form fields are not filled correctly. Consider incrementing the count when a new movie is added successfully to the list.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,100 @@
...
+    setCount(currentCount => currentCount + 1);
+
+    onAdd({
```
Viktor-Kost, line 32: You are adding a movie directly without any form validation. You should validate the form inputs before adding a new movie to the list.

Viktor-Kost (changes requested):
Hello!)
Please update the logic to avoid adding a films entering just spaces to the input
And don't allow to add movie when we have some errors
Screenshot 2023-06-30 at 21 09 24
Screenshot 2023-06-30 at 21 11 33

volodymyr-soltys97 (approved):

PR #1593, opened 2023-07-02
React components to add new movie were added

Olena-Shevchenko:
[DEMO LINK](https://Olena-Shevchenko.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,21 @@
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
 
 export const App = () => {
+  const [moviesList, setMoviesList] = useState([...moviesFromServer]);
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={moviesList} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie
+          onAdd={(newMovie) => setMoviesList([...moviesList, newMovie])}
+        />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,10 +1,46 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (newMovie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [titleValue, setTitleValue] = useState('');
+  const [descValue, setDescValue] = useState('');
+  const [imgUrlValue, setImgUrlValue] = useState('');
+  const [imdbUrlValue, setImdbUrlValue] = useState('');
+  const [imdbIdValue, setImdbIdValue] = useState('');
+
+  const submitDisabled = !titleValue || !imgUrlValue
+    || !imdbUrlValue || !imdbIdValue;
+
+  const clearForm = () => {
+    setTitleValue('');
+    setDescValue('');
+    setImgUrlValue('');
+    setImdbUrlValue('');
+    setImdbIdValue('');
+  };
+
+  const addNewMovie = (e: React.FormEvent) => {
+    const newMovie = {
+      title: titleValue,
+      description: descValue,
+      imgUrl: imgUrlValue,
+      imdbUrl: imdbUrlValue,
+      imdbId: imdbIdValue,
+    };
+
+    e.preventDefault();
+    onAdd(newMovie);
+    setCount(count + 1);
+    clearForm();
+  };
 
   return (
     <form className="NewMovie" key={count}>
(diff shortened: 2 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +91,8 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={submitDisabled}
+            onClick={addNewMovie}
```
lerastarynets, line 95: Pls use `onSubmit` handler for `form` tag instead of the `onClick` event on the submit button

src/App.tsx
```diff
@@ -1,16 +1,21 @@
...
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={moviesList} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie
+          onAdd={(newMovie) => setMoviesList([...moviesList, newMovie])}
```
lerastarynets, line 17: Just fyi, [better to handle such cases with prevState](https://blog.greenroots.info/react-hook-usestate-lazy-initialization-previous-state#heading-how-to-get-the-previous-state-data-and-use-it)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,46 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [titleValue, setTitleValue] = useState('');
+  const [descValue, setDescValue] = useState('');
+  const [imgUrlValue, setImgUrlValue] = useState('');
+  const [imdbUrlValue, setImdbUrlValue] = useState('');
+  const [imdbIdValue, setImdbIdValue] = useState('');
```
lerastarynets, line 17: Consider using 1 state for all the fields

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,46 @@
...
+
+  const submitDisabled = !titleValue || !imgUrlValue
+    || !imdbUrlValue || !imdbIdValue;
```
lerastarynets, line 20: Pls use `trim()` now i can fill and submit the form with whitespaces

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,46 @@
...
+      title: titleValue,
+      description: descValue,
+      imgUrl: imgUrlValue,
+      imdbUrl: imdbUrlValue,
+      imdbId: imdbIdValue,
+    };
+
+    e.preventDefault();
+    onAdd(newMovie);
+    setCount(count + 1);
```
lerastarynets, line 41: Better to use `prevState` here as well

src/components/TextField/TextField.tsx
```diff
@@ -43,7 +43,7 @@ export const TextField: React.FC<Props> = ({
           type="text"
           placeholder={`Enter ${label}`}
           value={value}
-          onChange={event => onChange(event.target.value)}
+          onChange={(e) => onChange(e.target.value)}
```
lerastarynets, line 46:
```suggestion
          onChange={(event) => onChange(event.target.value)}
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,46 @@
...
+
+  const clearForm = () => {
+    setTitleValue('');
+    setDescValue('');
+    setImgUrlValue('');
+    setImdbUrlValue('');
+    setImdbIdValue('');
+  };
+
+  const addNewMovie = (e: React.FormEvent) => {
```
lerastarynets, line 30:
```suggestion
  const addNewMovie = (event: React.FormEvent) => {
```

src/App.tsx
```diff
@@ -1,16 +1,21 @@
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
 
 export const App = () => {
+  const [moviesList, setMoviesList] = useState([...moviesFromServer]);
```
lerastarynets, line 8:
```suggestion
  const [moviesList, setMoviesList] = useState(moviesFromServer);
```
`moviesFromServer` is a constant, so can be used like so

src/App.tsx
lerastarynets, line 17: Also pls move the function inside `onAdd` handler to a separate const and use as a callback here

lerastarynets (changes requested):
Nice start! Let's fix the solution according to the checklist

src/components/NewMovie/NewMovie.tsx
Olena-Shevchenko: I don`t know why, but with onSubmit preventDefault() doesn`t work. That`s why I can`t change it

src/App.tsx
```diff
@@ -1,16 +1,27 @@
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+import { Movie } from './types/Movie';
 import moviesFromServer from './api/movies.json';
 
 export const App = () => {
+  const [moviesList, setMoviesList] = useState(moviesFromServer);
+  const onAdd = (newMovie: Movie) => {
```
polosanya, line 10:
```suggestion
  const addMovie = (newMovie: Movie) => {
```
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,60 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (newMovie: Movie) => void;
+};
+
+const clearMovieForm = {
```
polosanya, line 9: I think it's better to name it initial\empty instead of clear, as it can be confused with a verb

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,60 @@
...
+
+  const {
+    title,
+    description,
+    imgUrl,
+    imdbUrl,
+    imdbId,
+  } = movieForm;
+
+  const submitDisabled = !title.trim() || !imgUrl.trim()
```
polosanya, line 31:
```suggestion
  const isSubmitDisabled = !title.trim() || !imgUrl.trim()
```
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,60 @@
...
+      imdbUrl,
+      imdbId,
+    };
+
+    onAdd(newMovie);
+    setCount((prevCount) => prevCount + 1);
+    clearForm();
+  };
+
+  const onChangeHandler = (value: string, field: string) => {
```
polosanya, line 55:
```suggestion
  const handleChange = (value: string, field: string) => {
```

src/components/NewMovie/NewMovie.tsx
polosanya: Not fixed
image

polosanya (changes requested):
Almost done! Feel free to ask for help in fe_chat0 if you have any difficulties

src/components/NewMovie/NewMovie.tsx
Olena-Shevchenko: When I change onClick to on Submit my code doesn't work. I found that doesn't work preventDefault with on Submit.
I can't fixed it.

lerastarynets (approved):
All requirements are met. Keep up the great work!

PR #1595, opened 2023-07-03
added task solution

holytimy4:
[DEMO LINK](https://holytimy4.github.io/react_movies-list-add-form/)


roman-mirzoian (approved):
The job is well done and the application works correctly, I like the code that is written well and structured 👍

PR #1596, opened 2023-07-03
solution

MaxSchmide:
[DEMO LINK](https://maxschmide.github.io/react_movies-list-add-form/)


ollavka (approved):

anstsot (changes requested):
Found one bug to fix:
It's possible to add a movie with empty fields. (for tests I've typed a few spaces:   )
image

anstsot (approved):
approving:sparkles:
But it will be good to show errors as well

PR #1601, opened 2023-07-05
sol

GGsmou:
[DEMO LINK](https://GGsmou.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,23 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie: Movie) => {
+          setMovies(prevMovies => [...prevMovies, movie]);
+        }}
+        />
       </div>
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +90,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!title || !imgUrl || !imdbUrl || !imdbId}
```
adv555, line 93: the button is able if to put an empty value in inputs... might be needed to trim() values

adv555 (commented):
Nice work!✨
.. some small suggestions

adv555 (approved):

src/App.tsx
```diff
@@ -1,16 +1,23 @@
...
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie: Movie) => {
```
anstsot, line 17: better to create separate `onAdd` function
```suggestion
        <NewMovie onAdd={onAdd}
```

anstsot (changes requested):
Found one bug to fix:
It's possible to add a movie with empty fields. (for tests I've typed few spaces: `   `)
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +88,12 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={
+              !title.trim()
+              || !imgUrl.trim()
+              || !imdbUrl.trim()
+              || !imdbId.trim()
+            }
```
YegorKorenkov, line 96: Better move to separate variable to make jsx more cleaner

YegorKorenkov (approved):
Good job!

PR #1602, opened 2023-07-05
React_movies-list-add-form

ArtemKravcov85:
[DEMO LINK](https://ArtemKravcov85.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,99 @@
-import { useState } from 'react';
-import { TextField } from '../TextField';
+import { FC, useState } from 'react';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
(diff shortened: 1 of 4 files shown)


sTorba24 (changes requested):
Good work! Lets improve it

I can add just spaces, use trim to fix it


image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,102 @@
...
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    onAdd(formState);
+
+    clearForm();
+  };
+
+  const changeFormState = (key: string, value: string) => {
```
agniya-i, line 48: It would be better to start the function name with `handle`, same as `handleSubmit`

agniya-i (approved):

PR #1616, opened 2023-07-10
add task solution

ilona-borkovska:
[DEMO LINK](https://ilona-borkovska.github.io/react_movies-list-add-form/)


artkas03 (approved):
Everything is ok but i still can add a movie card even if one or both urls are invalid.
To avoid this you can disable submit button till you have invalid url too)

YegorKorenkov (changes requested):

anstsot (changes requested):
Found one bug to fix:
It's possible to add a movie with some empty fields. (for tests I've typed few spaces: `   `)
image

YegorKorenkov (approved):
Good job!

PR #1621, opened 2023-07-11
solution

vpdrabynko:
[DEMO LINK](https://vpdrabynko.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -8,6 +8,7 @@ type Props = {
   placeholder?: string,
   required?: boolean,
   onChange?: (newValue: string) => void,
+  pattern?: RegExp,
 };
 
 function getRandomDigits() {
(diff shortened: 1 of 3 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -22,14 +23,17 @@ export const TextField: React.FC<Props> = ({
   label = name,
   placeholder = `Enter ${label}`,
   required = false,
-  onChange = () => {},
+  onChange = () => { },
```
witflash, line 26: You don't really need the space between curly braces :)

witflash (changes requested):
Good work!
Just one thing. 
Could you validate your form fields with the `trim` method? Because a user can create an empty title for instance if he typed just spaces inside a field.

sTorba24 (approved):
Good work!

PR #1627, opened 2023-07-12
solution

DanyloPasko:
[DEMO LINK](https://DanyloPasko.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,126 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
+import { isValidUrl } from '../../test/validUrl';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (newMovie: Movie) => void;
+}
+
+const initialFormValues: Movie = {
+  title: '',
+  description: '',
+  imgUrl: '',
+  imdbUrl: '',
+  imdbId: '',
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [formValues, setFormValues] = useState(initialFormValues);
+  const [imgUrlError, setImgUrlError] = useState('');
+  const [imbdUrlError, setImbdUrlError] = useState('');
+
+  const {
+    title,
+    description,
+    imgUrl,
+    imdbUrl,
+    imdbId,
+  } = formValues;
+
+  const validValues = title.trim()
+    && imgUrl.trim()
+    && imdbUrl.trim()
+    && imdbId.trim();
+
+  const isDisabled = !validValues || !!imgUrlError || !!imbdUrlError;
+
+  const handleFormChange = (name: string, value: string) => {
+    if (name === 'imgUrl') {
+      if (!isValidUrl(value)) {
+        setImgUrlError('Invalid image URL');
+      } else {
+        setImgUrlError('');
+      }
+    }
+
+    if (name === 'imdbUrl') {
+      if (!isValidUrl(value)) {
+        setImbdUrlError('Invalid imbd URL');
+      } else {
+        setImbdUrlError('');
+      }
+    }
+
+    setFormValues({
+      ...formValues,
+      [name]: value,
+    });
+  };
+
+  const handleReset = () => {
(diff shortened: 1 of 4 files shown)


src/test/validUrl.ts
```diff
@@ -0,0 +1,6 @@
+export const isValidUrl = (value: string) => {
```
anastasiiavorobiova, line 1: There is no type annotation for the return value of the function. It is recommended to always annotate the return type of functions for better readability and error avoidance.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,126 @@
...
+      if (!isValidUrl(value)) {
+        setImgUrlError('Invalid image URL');
+      } else {
+        setImgUrlError('');
+      }
+    }
+
+    if (name === 'imdbUrl') {
+      if (!isValidUrl(value)) {
+        setImbdUrlError('Invalid imbd URL');
```
anastasiiavorobiova, line 50: The error message 'Invalid imbd URL' seems to have a typo. It should be 'Invalid imdb URL'.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,126 @@
...
+  };
+
+  const handleReset = () => {
+    setFormValues(initialFormValues);
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    handleReset();
```
anastasiiavorobiova, line 69: You should handle resetting the form after the new movie has been successfully added. In the current implementation, you are clearing the form values before the onAdd function is called. This can result in the addition of an empty movie object if onAdd is asynchronous and takes more time to execute.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,126 @@
...
+const initialFormValues: Movie = {
+  title: '',
+  description: '',
+  imgUrl: '',
+  imdbUrl: '',
+  imdbId: '',
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
```
anastasiiavorobiova, line 19: The 'count' state is initialized but it doesn't seem to be used anywhere in the component. If it's not used, consider removing it to avoid unnecessary state management.

src/components/TextField/TextField.tsx
```diff
@@ -22,14 +23,13 @@ export const TextField: React.FC<Props> = ({
   label = name,
   placeholder = `Enter ${label}`,
   required = false,
-  onChange = () => {},
+  onChange = () => { },
```
anastasiiavorobiova, line 26: There is an extra space between the arrow function brackets, this is not an error but it's considered as a bad practice.

src/components/TextField/TextField.tsx
```diff
@@ -22,14 +23,13 @@ export const TextField: React.FC<Props> = ({
...
+  errorMessage = '',
 }) => {
-  // generage a unique id once on component load
   const [id] = useState(() => `${name}-${getRandomDigits()}`);
 
-  // To show errors only if the field was touched (onBlur)
   const [touched, setTouched] = useState(false);
-  const hasError = touched && required && !value;
+  const hasError = (touched && required && !value);
```
anastasiiavorobiova, line 32: The additional parentheses around the boolean expression in 'hasError' are unnecessary and can be removed.

anastasiiavorobiova (changes requested):
Great start! Added a few comments to fix

src/components/NewMovie/NewMovie.tsx
DanyloPasko: counter used

src/test/validUrl.ts
```diff
@@ -0,0 +1,6 @@
+export const isValidUrl = (value: string): boolean => {
+  // eslint-disable-next-line max-len
+  const pattern = new RegExp(/^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/);
```
OleziO, line 3: Split regexp by a few line and remove rule disabling.
image

OleziO (changes requested):

anastasiiavorobiova (approved):
Great solution!

PR #1629, opened 2023-07-12
React__movies-list-add-form

Soi4An:
DEMO - https://Soi4An.github.io/react_movies-list-add-form/

Problems:
```
Page
  1) should add a movie with correct data
```
maybe becouse I add new movie on start array


vlad-demchuk (changes requested):
Your validation is working incorrectly due to I can't add an image with a correct link
image

I tested the regex, it is okay, so check your logic
image

volodymyr-soltys97 (approved):

PR #1642, opened 2023-07-17
add solution

IraShtynda:
[DEMO LINK](https://irashtynda.github.io/react_movies-list-add-form/)


anstsot (changes requested):
It will be good to show some errors in case of empty fields (by using `trim`)

anstsot (approved):
Good work, approving✨
It will be good to show some errors in case of empty fields (by using trim)

PR #1643, opened 2023-07-17
Solution

melnyksergey7:
https://melnyksergey7.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,89 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +92,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!title || !imgUrl || !imdbUrl || !imdbId}
```
MarianVolytskyi, line 95: it would be better to create new variable with checking.

MarianVolytskyi (changes requested):
Good job!
But, please make a little fix

anstsot (approved):

PR #1646, opened 2023-07-17
add task solution

dipinchuk:
[DEMO LINK](https://dipinchuk.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,58 +1,127 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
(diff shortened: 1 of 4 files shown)


eleonora-demikh (approved):
Well done!

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,58 +1,127 @@
...
+            type="submit"
+            data-cy="submit-button"
+            className="button is-link"
+            disabled
+          >
+            Add
+          </button>
+        </div>
+      </div>
+    );
```
anstsot, line 73: do not return 2 similar code snippets
```suggestion
    const isFormFilled = title && imgUrl && imdbUrl && imdbId;

    return (
      <div className="field is-grouped">
        <div className="control">
          <button
            type="submit"
            data-cy="submit-button"
            className="button is-link"
            disabled={!isFormFilled}
          >
            Add
          </button>
        </div>
      </div>
    );
```

anstsot (changes requested):
Found one bug to fix:
image

PR #1647, opened 2023-07-17
Develop

IgorOksentyuk:
https://igoroksentyuk.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,10 +1,56 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (newMovie: Movie) => void
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,56 @@
...
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [isButtonDisabled, setIsButtonDisabled] = useState(true);
+
+  const onTitleChange = (value: string) => {
+    setTitle(value);
+    setIsButtonDisabled(!value || !imgUrl || !imdbUrl || !imdbId);
```
outofs, line 20: There's no need of making separate state for `disabled` status of button and making necessary `setIsButtonDisabled`.
You can make just one constant, which will depends on title, imgUrl, imdUrl ... states.
`const isButtonDisabled = !title || !imgUrl || !imdbUrl || !imdbId`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,56 @@
...
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    });
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
```
outofs, line 50: Personally, I think it will be better if you put it in a separate 'reset' function.

outofs (approved):
Good work, but I have some advises below)

anstsot (changes requested):
Found one bug to fix:
image

anstsot (approved):

PR #1652, opened 2023-07-17
add task solution

ArtemAndrushchenko:
[DEMO LINK](https://ArtemAndrushchenko.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,90 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [descriptionField, setDescriptionField] = useState('');
+  const [titleField, setTitleField] = useState('');
+  const [imgUrlField, setImgUrlField] = useState('');
+  const [imdbUrlField, setImdbUrlField] = useState('');
+  const [imdbIdField, setImdbIdField] = useState('');
+
+  const handleSubmit = (event: React.FormEvent) => {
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,90 @@
...
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    const movieToUpload = {
+      title: titleField,
+      description: descriptionField,
+      imgUrl: imgUrlField,
+      imdbUrl: imdbUrlField,
+      imdbId: imdbIdField,
+    };
```
ovynarchuk, line 26: instead of writing too much code in your `movieToUpload` object, you could change names in your `useState` variables so it matches names in both cases
```suggestion
  const [count, setCount] = useState(0);
  const [description, setDescription] = useState('');
  const [title, setTitle] = useState('');
  const [imgUrl, setImgUrl] = useState('');
  const [imdbUrl, setImdbUrl] = useState('');
  const [imdbId, setImdbId] = useState('');

  const handleSubmit = (event: React.FormEvent) => {
    event.preventDefault();

    const movieToUpload = {
      title,
      description,
      imgUrl,
      imdbUrl,
      imdbId,
    };
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,90 @@
...
+
+    const reset = () => {
+      setDescriptionField('');
+      setTitleField('');
+      setImgUrlField('');
+      setImdbUrlField('');
+      setImdbIdField('');
+    };
```
ovynarchuk, line 34: and of course change them here and down further in your code accordingly

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,90 @@
...
+
+    onAdd(movieToUpload);
+    setCount((current) => current + 1);
```
ovynarchuk, line 37: here you could just write like this
```suggestion
    setCount(count + 1);
```

ovynarchuk (approved):
great job! I'm approving, but please check my comments down below

YegorKorenkov (changes requested):

anstsot (approved):

PR #1659, opened 2023-07-17
solution

artem-ivashchenko:
[DEMO LINK](https://artem-ivashchenko.github.io/react_movies-list-add-form/)


lerastarynets (approved):
You're showing incredible skills, keep it going!

PR #1663, opened 2023-07-17
add task solution

NadiaKovalchuk:
[DEMO LINK] https://NadiaKovalchuk.github.io/react_movies-list-add-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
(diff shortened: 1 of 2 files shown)


vlopatko (approved):
Good job looks great! 👍🏻

anstsot (changes requested):
Found one bug to fix:
image

And678 (changes requested):
Entered spaces, button is disabled, yet no error messages 
image

NadiaKovalchuk:
> Entered spaces, button is disabled, yet no error messages image

Hello) thank you for your feedback, but the demo version works like mine, now you can't enter empty spaces it and add them, because the button doesn't work, but any mistakes there after you enter open spaces I can`t see. So do I need to do some improvements or can I live it that way?

anstsot (approved):

PR #1668, opened 2023-07-17
solution

Valentyna08:
[DEMO LINK](https://Valentyna08.github.io/react_movies-list-add-form/)


anstsot (changes requested):
Found one bug to fix:
image

anstsot (approved):
But it will be good to show different errors

PR #1686, opened 2023-07-19
Form was added to the movies list

serge-kontakt:
[DEMO LINK](https://serge-kontakt.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,112 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [validImg, setValidImg] = useState(true);
+  const [validImdb, setValidImdb] = useState(true);
+
+  function handleTextField(setText: (value: string) => void) {
+    const handleText = (value: string) => {
+      setText(value);
+    };
+
+    return handleText;
+  }
+
+  let disabledButton = true;
+
+  if (title && imgUrl && imdbUrl && imdbId && validImg === true
+    && validImdb === true) {
+    disabledButton = false;
+  }
+
+  const handleSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
+
+    onAdd({
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    });
+
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+    setValidImg(true);
+    setValidImdb(true);
+    setCount((countPrev) => countPrev + 1);
+  };
 
   return (
-    <form className="NewMovie" key={count}>
+    <form
+      className="NewMovie"
+      key={count}
+      onSubmit={handleSubmit}
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        setValidImg={setValidImg}
+        setValidImdb={setValidImdb}
+        onChange={handleTextField(setTitle)}
         required
       />
 
       <TextField
         name="description"
         label="Description"
-        value=""
+        value={description}
+        setValidImg={setValidImg}
+        setValidImdb={setValidImdb}
+        onChange={handleTextField(setDescription)}
       />
 
       <TextField
         name="imgUrl"
         label="Image URL"
-        value=""
+        value={imgUrl}
+        setValidImg={setValidImg}
+        setValidImdb={setValidImdb}
+        onChange={handleTextField(setImgUrl)}
+        required
       />
 
       <TextField
         name="imdbUrl"
         label="Imdb URL"
-        value=""
+        value={imdbUrl}
+        setValidImg={setValidImg}
+        setValidImdb={setValidImdb}
+        onChange={handleTextField(setImdbUrl)}
+        required
       />
 
       <TextField
         name="imdbId"
         label="Imdb ID"
-        value=""
+        value={imdbId}
+        setValidImg={setValidImg}
+        setValidImdb={setValidImdb}
+        onChange={handleTextField(setImdbId)}
+        required
       />
 
       <div className="field is-grouped">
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,112 @@
...
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [validImg, setValidImg] = useState(true);
```
polosanya, line 16:
```suggestion
  const [isValidImg, setIsValidImg] = useState(true);
```
image
https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,112 @@
...
+    };
+
+    return handleText;
+  }
+
+  let disabledButton = true;
+
+  if (title && imgUrl && imdbUrl && imdbId && validImg === true
+    && validImdb === true) {
+    disabledButton = false;
```
polosanya, line 31:
```suggestion
  const canSubmit = title && imgUrl && imdbUrl && imdbId && validImg
    && validImdb;
```
Just save the condition as variable. Also, there is no need to compare the boolean value with true or false. Use boolean/!boolean instead

polosanya (changes requested):
Almost perfect!

src/components/NewMovie/NewMovie.tsx
serge-kontakt: There is a TS Error if I change condition to variable.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,111 @@
...
+    };
+
+    return handleText;
+  }
+
+  let cantSubmit = true;
+
+  if (title && imgUrl && imdbUrl && imdbId && isValidImg && isValidImdb) {
+    cantSubmit = false;
+  }
```
Anastasiia-Svintsova, line 31:
```suggestion
  const canSubmit = !!(title && imgUrl && imdbUrl && imdbId && isValidImg && isValidImdb);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,111 @@
...
+    });
+
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+    setIsValidImg(true);
+    setIsValidImdb(true);
+    setCount((countPrev) => countPrev + 1);
```
Anastasiia-Svintsova, line 51: For what purposes is a count needed here?

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,111 @@
...
+  const [isValidImg, setIsValidImg] = useState(true);
+  const [isValidImdb, setIsValidImdb] = useState(true);
+
+  function handleTextField(setText: (value: string) => void) {
+    const handleText = (value: string) => {
+      setText(value);
+    };
+
+    return handleText;
+  }
```
Anastasiia-Svintsova, line 25: You can simplify it
```suggestion
  function handleTextField(setText: (value: string) => void) {
    setText(value);
  }
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,111 @@
...
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        setIsValidImg={setIsValidImg}
+        setIsValidImdb={setIsValidImdb}
+        onChange={handleTextField(setTitle)}
```
Anastasiia-Svintsova, line 68: It directly invokes the handleTextField function when the component renders. The onChange event expects a function reference that will be called when the input's value changes. 
```
onChange={() => ....}

**Review by Anastasiia-Svintsova — CHANGES_REQUESTED** [review]
- The user shouldn't be able to add a movie with empty fields
image

- Validation should use the trim() method to check if the input is not empty

https://github.com/mate-academy/react_movies-list-add-form/assets/99403198/b7bbedd1-38e8-47e8-95d0-1a7821a5473b

- Use a trim() method before validating an URL
image

File: src/components/NewMovie/NewMovie.tsx (hunk 4 above)
Comment (serge-kontakt) [reply src/components/NewMovie/NewMovie.tsx:51]: it was in the comments in master: 

File: src/components/NewMovie/NewMovie.tsx (hunk 5 above)
Comment (serge-kontakt) [reply src/components/NewMovie/NewMovie.tsx:25]: it does not work - "Cannot find name 'value'"

Comment (Anastasiia-Svintsova) [reply src/components/NewMovie/NewMovie.tsx:25]: Sorry, my bad. You can pass the value as the argument as well

File: src/components/NewMovie/NewMovie.tsx (hunk 6 above)
Comment (Anastasiia-Svintsova) [reply src/components/NewMovie/NewMovie.tsx:67]: Not fixed. this function shouldn't be invoked on every render. BTW You can simplify it and get rid of the handleTextField method. 
```suggestion
        onChange={setTitle}
```

src/components/TextField/TextField.tsx
```diff
@@ -23,13 +25,51 @@ export const TextField: React.FC<Props> = ({
...
+      setIsValidImdb(false);
+    }
+
+    if (isValidInput && newValue && name === 'imgUrl') {
+      setIsValidImg(true);
+    }
+
+    if (isValidInput && newValue && name === 'imdbUrl') {
+      setIsValidImdb(true);
+    }
```
Anastasiia-Svintsova, line 69: You can remove redundant checks.
```suggestion
    if (name === 'imgUrl') {
      setIsValidImg(isValidInput);
    }

    if (name === 'imdbUrl') {
      setIsValidImdb(isValidInput);
    }
```

src/components/TextField/TextField.tsx
```diff
@@ -23,13 +25,51 @@ export const TextField: React.FC<Props> = ({
...
+
 }) => {
-  // generage a unique id once on component load
   const [id] = useState(() => `${name}-${getRandomDigits()}`);
 
-  // To show errors only if the field was touched (onBlur)
   const [touched, setTouched] = useState(false);
-  const hasError = touched && required && !value;
+  const hasError = touched && required && !value.trim();
+  let hasUrlError = false;
```
Anastasiia-Svintsova, line 36: It's better to use setState for this purposes

Anastasiia-Svintsova (changes requested):
URL validation still doesn't work correctly. Ask questions in the chat if you need some help
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,102 @@
...
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [isValidImg, setIsValidImg] = useState(true);
+  const [isValidImdb, setIsValidImdb] = useState(true);
+
+  const cantSubmit = (
```
agniya-i, line 19: Better to prefix boolean variables with `has` or `is`,  you can rename it to `isValid`, `isButtonEnabled` etc

agniya-i (approved):
👍

PR #1690, opened 2023-07-21
solution

diana-tuz:
[DEMO LINK](https://diana-tuz.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,87 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isDisabled = () => {
+    if (!title || !imdbId || !imdbUrl || !imgUrl) {
+      return true;
+    }
+
+    return false;
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (title && imdbId && imdbUrl && imgUrl) {
+      onAdd({
+        title,
+        description,
+        imgUrl,
+        imdbUrl,
+        imdbId,
+      });
+      setCount(count + 1);
+      setTitle('');
+      setDescription('');
+      setImgUrl('');
+      setImdbUrl('');
+      setImdbId('');
+    }
+  };
 
   return (
-    <form className="NewMovie" key={count}>
+    <form className="NewMovie" key={count} onSubmit={handleSubmit}>
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(newValue) => setTitle(newValue)}
         required
       />
 
       <TextField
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,87 @@
...
+
+  const isDisabled = () => {
+    if (!title || !imdbId || !imdbUrl || !imgUrl) {
+      return true;
+    }
+
+    return false;
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
```
anastasiiavorobiova, line 25: The task requires validation for the image URL and IMDB URL fields using a regex pattern. This is currently not implemented.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,87 @@
...
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isDisabled = () => {
```
anastasiiavorobiova, line 17: Consider simplifying the 'isDisabled' function by directly returning the result of the logical expression.

src/components/NewMovie/NewMovie.tsx
anastasiiavorobiova, line 25: The task requires form validation on blur event for each input field. This is currently not implemented.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,87 @@
...
+    if (!title || !imdbId || !imdbUrl || !imgUrl) {
+      return true;
+    }
+
+    return false;
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (title && imdbId && imdbUrl && imgUrl) {
```
anastasiiavorobiova, line 27: You are repeating the same checks in the 'handleSubmit' function that you have already done in the 'isDisabled' function. You can remove these checks from 'handleSubmit'.

src/components/TextField/TextField.tsx
```diff
@@ -17,12 +17,12 @@ function getRandomDigits() {
 }
 
 export const TextField: React.FC<Props> = ({
-  name,
-  value,
+  name = '',
+  value = '',
```
anastasiiavorobiova, line 21: These fields shouldn't be empty strings. Name and value are mandatory values

anastasiiavorobiova (changes requested):
Great work! Added comments to fix. I was able to add the movie using spaces for every field. Consider implementing validation and trim values before sending
Screenshot 2023-07-21 at 19 00 34

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+        value={description || ''}
+        onChange={(newValue) => setDescription(newValue)}
       />
 
       <TextField
         name="imgUrl"
         label="Image URL"
-        value=""
+        value={imgUrl}
+        onChange={(newValue) => setImgUrl(newValue.trim())}
```
volodymyr-soltys97, line 73: You are using trim() function while setting the imgUrl state. This might not be necessary and could lead to incorrect URLs. It would be more appropriate to trim the URL before checking it against the regex pattern during validation.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+        validate={(value) => pattern.test(value)}
+        required
       />
 
       <TextField
         name="imdbUrl"
         label="Imdb URL"
-        value=""
+        value={imdbUrl}
+        onChange={(newValue) => setImdbUrl(newValue.trim())}
```
volodymyr-soltys97, line 82: You are using trim() function while setting the imdbUrl state. This might not be necessary and could lead to incorrect URLs. It would be more appropriate to trim the URL before checking it against the regex pattern during validation.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

volodymyr-soltys97 (approved):
Well done 🔥

PR #1693, opened 2023-07-23
done

vetal-hovenko:
[demo](https://vetal-hovenko.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,106 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imageUrl, setImageUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const fieldsAreFilled = title && imageUrl && imdbUrl && imdbId;
+
+  const buttonDisabling = () => {
+    if (fieldsAreFilled) {
+      return false;
+    }
+
+    return true;
+  };
+
+  const formSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (fieldsAreFilled) {
+      const newMovie: Movie = {
+        title,
+        description,
+        imgUrl: imageUrl,
+        imdbUrl,
+        imdbId,
+      };
+
+      onAdd(newMovie);
+
+      setTitle('');
+      setDescription('');
+      setImageUrl('');
+      setImdbUrl('');
+      setImdbId('');
+      setCount((prevCount) => prevCount + 1);
+    }
+  };
 
   return (
-    <form className="NewMovie" key={count}>
+    <form
+      className="NewMovie"
+      key={count}
+      onSubmit={(event) => formSubmit(event)}
+    >
       <h2 className="title">Add a movie</h2>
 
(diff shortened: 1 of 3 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
anastasiiavorobiova, line 1: You should avoid disabling eslint rules, unless it is absolutely necessary. If you have unused variables, it's better to remove them.

src/components/TextField/TextField.tsx
```diff
@@ -29,7 +30,8 @@ export const TextField: React.FC<Props> = ({
 
   // To show errors only if the field was touched (onBlur)
   const [touched, setTouched] = useState(false);
-  const hasError = touched && required && !value;
+
+  const hasError = (touched && (required && !value));
```
anastasiiavorobiova, line 34: There is no need to wrap the second part of the condition in additional parentheses. It does not affect the logic and can be removed for better readability.

src/components/TextField/TextField.tsx
```diff
@@ -49,6 +51,7 @@ export const TextField: React.FC<Props> = ({
           value={value}
           onChange={event => onChange(event.target.value)}
           onBlur={() => setTouched(true)}
+          onSubmit={() => setTouched(true)}
```
anastasiiavorobiova, line 54: The onSubmit event is not applicable to input elements. It should be used on form elements. If you want to handle the form submission, it should be done in the form component, not in the individual input component.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,106 @@
...
+        onChange={(event) => setDescription(event)}
       />
 
       <TextField
         name="imgUrl"
         label="Image URL"
-        value=""
+        value={imageUrl}
+        onChange={(event) => {
+          setImageUrl(event);
```
anastasiiavorobiova, line 81: The event object is being passed directly to the setImageUrl function which is incorrect. You should pass event.target.value instead.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,106 @@
...
+        }}
         required
       />
 
       <TextField
         name="description"
         label="Description"
-        value=""
+        value={description}
+        onChange={(event) => setDescription(event)}
```
anastasiiavorobiova, line 73: The event object is being passed directly to the setDescription function which is incorrect. You should pass event.target.value instead.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,106 @@
...
+        required
       />
 
       <TextField
         name="imdbUrl"
         label="Imdb URL"
-        value=""
+        value={imdbUrl}
+        onChange={(event) => {
+          setImdbUrl(event);
```
anastasiiavorobiova, line 91: The event object is being passed directly to the setImdbUrl function which is incorrect. You should pass event.target.value instead.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,106 @@
...
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(event) => {
+          setTitle(event);
```
anastasiiavorobiova, line 64: The event object is being passed directly to the setTitle function which is incorrect. You should pass event.target.value instead.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,106 @@
...
+        required
       />
 
       <TextField
         name="imdbId"
         label="Imdb ID"
-        value=""
+        value={imdbId}
+        onChange={(event) => {
+          setImdbId(event);
```
anastasiiavorobiova, line 101: The event object is being passed directly to the setImdbId function which is incorrect. You should pass event.target.value instead.

anastasiiavorobiova (changes requested):
Great start! Consider fixing comments and implementing validations. I was able to submit the form with spaces only
Screenshot 2023-07-23 at 17 09 01
A user shouldn't be able to do it.

volodymyr-soltys97 (approved):
Good job 👍

PR #1702, opened 2023-07-27
movies-list-add-form

Tetiana-Hishchak:
https://Tetiana-Hishchak.github.io/react_movies-list-add-form/


src/App.tsx
```diff
@@ -1,16 +1,23 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movie, setMovie] = useState(moviesFromServer);
```
Viktor-Kost, line 9: The variable name 'movie' is misleading as it actually represents an array of movies. Consider renaming it to 'movies' for better readability and understanding of the code.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,98 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [movie, setMovie] = useState<Movie>(initialMovieState);
+  const handleMovieChange = (event: React.ChangeEvent<HTMLInputElement>,
+    key: string) => {
+    setMovie(prevMovie => ({ ...prevMovie, [key]: event.target.value }));
+  };
```
Viktor-Kost, line 24: In the 'handleMovieChange' function, the 'key' argument is of type string. This can be risky because if the key does not exist in the movie object, it will create an error. To prevent this, you should use `keyof Movie` for the key parameter.

```suggestion
  const handleMovieChange = (event: React.ChangeEvent<HTMLInputElement>, key: keyof Movie) => {
    setMovie(prevMovie => ({ ...prevMovie, [key]: event.target.value }));
  };
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,98 @@
...
+    || !newMovie.imdbId.trim() || !newMovie.imdbUrl.trim()) {
+      return false;
+    }
+
+    if ((!newMovie.imgUrl.match(pattern) && newMovie.imgUrl !== '')
+    || (!newMovie.imdbUrl.match(pattern) && newMovie.imdbUrl !== '')) {
+      return false;
+    }
+
+    return true;
```
Viktor-Kost, line 37: That's what we're talking about. Nice job, since we can even simplify this condition
```suggestion
   const { title, imgUrl, imdbId, imdbUrl } = newMovie;
  
  return (
    title.trim() &&
    imgUrl.trim() &&
    imdbId.trim() &&
    imdbUrl.trim() &&
    imgUrl.match(pattern) &&
    imdbUrl.match(pattern)
  );
```

src/components/Service/RegularFunction.ts
```diff
@@ -0,0 +1,11 @@
+export const pattern = new RegExp(
```
Viktor-Kost, line 1: it's better to create a folder `constants` at the same level as `components` folders and give `patterns.ts` name to this file since this is not a function and not a service

Viktor-Kost (changes requested):
Great!
Let's improve your solution a little

VitaliyBondarenko1982 (approved):

PR #1703, opened 2023-07-28
Solution

artyomwhite:
[DEMO LINK](https://artyomwhite.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,114 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,114 @@
...
+    return handleText;
+  }
+
+  const isDiabled = () => {
+    if (title && imgUrl && imdbUrl && imdbId && validImg === true
+      && validImdb === true) {
+      return false;
+    }
+
+    return true;
```
VitaliyBondarenko1982, line 33: fix typo `isDisabled'
also redundant to make function for it.
just make variable isDisabled which will be set to true if one of fields invalid (use `||`)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,114 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
VitaliyBondarenko1982, line 15: make one state for all form fields and change proper field with event.target.name
```suggestion
  const [movie, setMovie] = useState({
    title: '',
    description: '',
    imgUrl: '',
    imdbUrl: '',
     imdbId: '',
   })
```
initial state better to write in constant outside of component and use here and for reset form

VitaliyBondarenko1982 (changes requested):
now i can add film where title, description just string with spaces
use trim() to fix it
Screenshot from 2023-07-28 13-25-53
Screenshot from 2023-07-28 13-15-51

PR #1706, opened 2023-07-29
Solution

artyomwhite:
[DEMO LINK](https://artyomwhite.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -2,87 +2,81 @@ import { useState } from 'react';
 import { TextField } from '../TextField';
 import { Movie } from '../../types/Movie';
 
+const initialMovieState: Movie = {
+  title: '',
+  description: '',
+  imgUrl: '',
+  imdbUrl: '',
+  imdbId: '',
+};
+
 type Props = {
   onAdd: (movie: Movie) => void;
 };
 
 export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   const [count, setCount] = useState(0);
-  const [title, setTitle] = useState('');
-  const [description, setDescription] = useState('');
-  const [imgUrl, setImgUrl] = useState('');
-  const [imdbUrl, setImdbUrl] = useState('');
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,95 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
VitaliyBondarenko1982, line 15: make one state for all form fields and change proper field with event.target.name
initialValue better to write in constant outside of components and use here and when reset form.
```
   title: '',
   description: '',
   ...
})
```

VitaliyBondarenko1982 (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,89 @@
...
+        onChange={(value) => handleChange('title', value)}
         required
       />
 
       <TextField
         name="description"
         label="Description"
-        value=""
+        value={movie.description}
+        onChange={(value) => handleChange('description', value)}
```
VitaliyBondarenko1982, line 56: no need to hardcode each field.
just pass reference on handleChange to TextField component
```suggestion
        onChange={handleChange}
```
also in TextFiled component change type of onChange prop
```
interface Props {
...
onChange: (event: React.ChangeEvent<HtmlInputElement>) => void;
}
```
and to input pass name and onChange reference
```
<input
  name={name}
  onChange={onChange}
  ...
 >
 ```

VitaliyBondarenko1982 (changes requested):
Needs some additional refactoring
Please check comments.

PR #1712, opened 2023-07-31
task done

tecHamster:
[DEMO LINK](https://tecHamster.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [preparedMovies, setPreparedMovies] = useState(moviesFromServer);
+
+  const onAdd = (movie: Movie) => {
+    setPreparedMovies(prevMovies => [...prevMovies, movie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={preparedMovies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie: Movie) => onAdd(movie)} />
       </div>
     </div>
   );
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={preparedMovies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie: Movie) => onAdd(movie)} />
```
lerastarynets, line 21: You don't need to create a new anonymous function in the onAdd prop. You can just pass the onAdd function directly.

lerastarynets (changes requested):
Unfortunately, your project is not reachable via the demo link(

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,110 @@
...
+      imdbId,
+    });
+
+    reset();
+  };
+
+  const [isValidImgUrl, setIsValidImgUrl] = useState(false);
+  const [isValidImdbUrl, setIsValidImdbUrl] = useState(false);
+
+  const hasURLErrors = isValidImgUrl || isValidImdbUrl;
```
lerastarynets, line 46: The variable 'hasURLErrors' will be 'true' if either 'isValidImgUrl' or 'isValidImdbUrl' is 'true'. This seems incorrect because you should show errors when the URLs are invalid, not valid. Consider using '!isValidImgUrl || !isValidImdbUrl' instead.
Maybe you can change namings

lerastarynets (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,110 @@
...
+      imdbId,
+    });
+
+    reset();
+  };
+
+  const [isValidImgUrl, setIsValidImgUrl] = useState(false);
+  const [isValidImdbUrl, setIsValidImdbUrl] = useState(false);
+
+  const hasURLErrors = !isValidImgUrl || !isValidImdbUrl;
```
maxim2310, line 46: consider the possibility to move these all const to other variables

maxim2310 (approved):
Great job!!!

PR #1713, opened 2023-08-01
solution

anna-tovkailo:
https://anna-tovkailo.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,101 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
+import { pattern } from '../../utils';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,101 @@
...
+    });
+  };
+
+  const formIsReady = (movie.title) && movie.imdbUrl.match(pattern)
+    && movie.imgUrl.match(pattern) && movie.imdbId.length > 0;
+
+  const handleInputField = (name: string, event: string) => {
+    setMovie(currentCount => ({
+      ...currentCount,
+      [name]: event.trim(),
```
TarasHoliuk, line 40: what if the movie name includes spaces? trim value when you check whether it's empty, not here

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,101 @@
...
+        onChange={(event) => {
+          handleInputField('description', event);
+        }}
       />
 
       <TextField
         name="imgUrl"
         label="Image URL"
-        value=""
+        value={movie.imgUrl}
```
TarasHoliuk, line 74: show a special error message if URL isn't valid:
image

TarasHoliuk (changes requested):

PR #1715, opened 2023-08-04
Solution

Pa1eOrc:

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,28 @@
+import { useState } from 'react';
+
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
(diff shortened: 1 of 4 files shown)


etojeDenys (changes requested):
let's disable form submitting only with spaces using [trim](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/String/trim)

src/App.tsx
```diff
@@ -1,16 +1,28 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [
+    copyMoviesFromServer,
+    setMoviesFromServer,
```
SanyaBratashchuk, line 12:
```suggestion
    movies,
    setMovies
```

SanyaBratashchuk (approved):

PR #1718, opened 2023-08-07
add task solution

vovaAlieksieiev:
[DEMO LINK](https://vovaAlieksieiev.github.io/react_movies-list-add-form/)


volodymyr-soltys97 (changes requested):
There is no way to add a movie because it keeps saying that the address is wrong, try adding it yourself and see if it works
image

PR #1719, opened 2023-08-08
Solution

s3rg4sh:
https://s3rg4sh.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,98 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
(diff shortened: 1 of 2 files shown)


VitaliyBondarenko1982 (changes requested):
Good job.
Needs some improvements.
Also check such behavior.
As user I can input all spaces to each form field but any error visible and submit button still disabled.
chrome-capture-2023-7-8

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,102 @@
...
+      onSubmit={handleSubmit}
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        movie={movie}
```
VitaliyBondarenko1982, line 61:
```suggestion
```
redundant to pass movie. use callback in setMovie function for change prev state

src/components/TextField/TextField.tsx
```diff
@@ -41,13 +44,16 @@ export const TextField: React.FC<Props> = ({
...
           data-cy={`movie-${name}`}
           className={classNames('input', {
             'is-danger': hasError,
           })}
           placeholder={placeholder}
           value={value}
-          onChange={event => onChange(event.target.value)}
+          onChange={event => onChange(
+            { ...movie, [event.target.name]: event.target.value },
+          )}
```
VitaliyBondarenko1982, line 56: it is bad practice to make anonymous arrow functions in jsx.
instead make handler in jsx and pass reference to jsx.
destructure event.target.
use prevMovie state.
```
const handleChange = (event: React.ChangeEvent<HtmlInputElement>) => {
   onChange( prevMovie =>  ({ ...prevMovie, [name]: value, });
}
```

```suggestion
          onChange={handleChange}
```

VitaliyBondarenko1982 (changes requested):
Almost done.
Check a couple comments for some improves.

IvanFesenko (approved):
Please check my comment to avoid same mistake in the future.

It will works same, but one callback less

PR #1720, opened 2023-08-08
solution

HYPER2307:
[DEMO LINK](https://hyper2307.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,50 +1,95 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({
+  onAdd,
+}) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
+
   const [count] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isValid = title && imgUrl && imdbId && imdbUrl;
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!isValid) {
+      return;
+    }
+
+    setTitle('');
+    setDescription('');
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,95 @@
...
+}) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
+
   const [count] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
etojeDenys, line 20: let's combine it in one useState

```
   title: '',
   ...
})

Comment (etojeDenys) [reply src/components/NewMovie/NewMovie.tsx:20]: and don't forget to rewrite the handle function 
```
 const {name,value} = e.target;
 setNewMovie({...newMovie, [name]: value})

}
```

Comment (etojeDenys) [reply src/components/NewMovie/NewMovie.tsx:20]: also don't forget to check if input has name attr

**Review by etojeDenys — CHANGES_REQUESTED** [review]
i also can add form only with spaces, let's fix that using trim

File: src/components/NewMovie/NewMovie.tsx (hunk 1 above)
Comment (HYPER2307) [reply src/components/NewMovie/NewMovie.tsx]: get it

File: src/components/TextField/TextField.tsx [hunk 2]
```diff
@@ -47,7 +48,7 @@ export const TextField: React.FC<Props> = ({
           })}
           placeholder={placeholder}
           value={value}
-          onChange={event => onChange(event.target.value)}
+          onChange={event => onChange(event)}
```
IvanFesenko, line 51: Will works same but less callbacks
```suggestion
          onChange={onChange}
```

IvanFesenko (changes requested):
Fix this case
https://github.com/mate-academy/react_movies-list-add-form/assets/61904995/b077fb3e-2388-4b02-aa92-e88575a477e3

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,113 @@
...
+      imdbId: '',
+    });
+
+    onAdd({
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    });
```
etojeDenys, line 61:
```suggestion
    onAdd(newMovie);
```

etojeDenys (changes requested):
nit

VitaliyBondarenko1982 (approved):
Looks good

PR #1723, opened 2023-08-09
Movie list add

ValeraViachalo:
[DEMO LINK](https://ValeraViachalo.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,97 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [movie, setMovie] = useState<Movie>({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+
+  const resetForm = () => {
+    setCount(count + 1);
+
+    setMovie({
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
+  };
+
+  const isFilled = (
+    !movie.title.trim()
+    || !movie.imgUrl.trim()
+    || !movie.imdbUrl.trim()
+    || !movie.imdbId.trim()
+  );
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (isFilled) {
+      return;
+    }
+
+    onAdd(movie);
+
+    resetForm();
+  };
 
   return (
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,97 @@
...
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={movie.title}
+        onChange={newMovie => setMovie(newMovie)}
```
Viktor-Kost, line 62: The onChange function is incorrect. It is currently replacing the whole movie object with the new value. It should instead update the specific property of the movie object that has changed. Implement a function to handle this.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,97 @@
...
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
+  };
+
+  const isFilled = (
+    !movie.title.trim()
```
Viktor-Kost, line 32: The isFilled condition is wrong. It should return true if all fields are filled, but it currently returns true if any field is not filled. Correct the logic, it should be '&&' instead of '||'.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,97 @@
...
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [movie, setMovie] = useState<Movie>({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
```
Viktor-Kost, line 17: move the state object to a variable outside the component, so you can reuse this object for reset function

```
const initialMovieState = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: ''
}
```

```suggestion
  const [movie, setMovie] = useState<Movie>(initialMovieState);
```

Viktor-Kost (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,109 @@
...
+
+  const [count, setCount] = useState(0);
+  const [movie, setMovie] = useState<Movie>(EMPTY_MOVIE);
+
+  const resetForm = () => {
+    setCount(count + 1);
+    setMovie(EMPTY_MOVIE);
+  };
+
+  const isFilled = !(
```
Viktor-Kost, line 26: Check if imgUrl and imdbUrl are valid URLs (you can use the regex)
Create this variable outside the component
```
const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;

```

Viktor-Kost (changes requested):
Nice!
Some minor fixes needed

Viktor-Kost (changes requested):
.

IvanFesenko (approved):
GJ

PR #1732, opened 2023-08-15
Solution

SashaViun:
[DEMO LINK](https://SashaViun.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,97 @@
...
+
+    setCount(count + 1);
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
+
+  const validTitle = (newValue: string) => {
```
anastasiiavorobiova, line 48: The function 'validTitle' is redundant. It does not add any extra functionality beyond directly calling 'setTitle'. You can simply use 'setTitle' instead.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,97 @@
...
+        validate={(value) => pattern.test(value)}
+        required
       />
 
       <TextField
         name="imdbId"
         label="Imdb ID"
-        value=""
+        value={imdbId}
+        onChange={(newValue) => setImdbId(newValue.trim())}
```
anastasiiavorobiova, line 93: Trimming the 'imdbId' field can be moved to the 'handleSubmit' function. It is better to do all the sanitization and trimming at once before the data is used or sent to the server.

anastasiiavorobiova (changes requested):
Amazing work! Unfortunately, I was able to create a film with an empty title and description. Consider trimming values for it
Screenshot 2023-08-15 at 18 29 11

volodymyr-soltys97 (approved):

PR #1738, opened 2023-08-16
Solution

RomanHrytsiuk:
https://RomanHrytsiuk.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -53,7 +53,7 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
       return;
     }
 
-    onAdd(movie);
+    onAdd(newMovie);
 
     reset();
 


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,101 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```

```
  title: '',
  ...
})

Comment (etojeDenys) [reply src/components/NewMovie/NewMovie.tsx:15]: now we can also create only one handle function for all fields
```
}
```

Comment (etojeDenys) [reply src/components/NewMovie/NewMovie.tsx:15]: don't forget to specify `name` attr to input in TextField

**Review by etojeDenys — CHANGES_REQUESTED** [review]

File: src/components/NewMovie/NewMovie.tsx [hunk 2]
```diff
@@ -1,45 +1,113 @@
...
+  };
+
+  function handleSubmit(event: React.FormEvent) {
+    event.preventDefault();
+
+    if (!emptyFields) {
+      return;
+    }
+
+    onAdd(movie);
```
etojeDenys, line 56:
```suggestion
    onAdd(newMovie);
```
i think it should be newMovie state instead of the empty object

etojeDenys (changes requested):
nit

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,113 @@
...
+    }));
+  };
+
+  const movie = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
```
etojeDenys, line 37: by the way, we can move this object out of the component

etojeDenys (approved):
lgtm

PR #1741, opened 2023-08-19
added solution

constantweb:
[DEMO LINK](https://constantweb.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,89 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type MovieData = {
+  title: string;
+  description: string;
+  imgUrl: string;
+  imdbUrl: string;
+  imdbId: string;
+};
+
+type Props = {
+  onAdd: (movie: MovieData) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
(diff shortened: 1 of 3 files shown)


maxim2310 (changes requested):

1. firstly must be required error 
image
2. error message must hide when i enter value
image

lerastarynets (changes requested):
the user is now able to add item with title consisting of white spaces only, pls fix

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,92 @@
...
+  onAdd: (movie: MovieData) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,93 @@
...
+  const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    onAdd(newMovie);
+    setNewMovie({
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
```
anastasiia-tilikina, line 35: let's create a variable with initial state and use it instead of duplicating the same obj twise

anastasiia-tilikina (changes requested):
almost done!
just one more minor fix is needed

lerastarynets (approved):
You're making excellent progress!

maxim2310 (approved):
Good work!!!

PR #1742, opened 2023-08-19
added disabled-eslint

Ilya-Osoben:
[DEMO LINK](https://Ilya-Osobenn.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movie, setMovie] = useState(moviesFromServer);
+
+  const addMovie = (movieNew: Movie) => {
+    setMovie((currentMovies) => [...currentMovies, movieNew]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movie} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addMovie} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,54 +1,109 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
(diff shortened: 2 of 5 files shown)


maxim2310 (changes requested):
Yor demo link is not working, please fix it and make request review again

src/App.tsx
```diff
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movie, setMovie] = useState(moviesFromServer);
```
lerastarynets, line 9: The variable name 'movie' is misleading for an array of movies. It would be more appropriate to use 'movies' or similar to depict that it's a collection.

src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
+
+  const addMovie = (movieNew: Movie) => {
```
lerastarynets, line 11: The parameter name 'movieNew' could be improved for readability. Something like 'newMovie' might be more clear.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,54 +1,109 @@
...
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(field) => setTitle(field)}
```
lerastarynets, line 67: The onChange event handlers for the TextField components are not being used properly. You should extract the event target value from the event object, not just passing the event object to the state update function.
Same for all textfields on the page

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,54 +1,109 @@
...
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  /* eslint-disable */
+  const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
+  /* eslint-enable */
+
+  const emptyFields = title.trim() && imgUrl.match(pattern)
```
lerastarynets, line 23: The variable name 'emptyFields' is misleading as it seems to indicate that the fields are empty, but it actually checks if the fields are not empty and valid. Consider renaming it to 'validFields' or something similar.

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

PR #1743, opened 2023-08-20
Movies List Add form

dimkamg21:
[DEMO LINK](https://dimkamg21.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,96 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,96 @@
...
+
+  const resetForm = () => {
+    setNewMovie({
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
+  };
```
VitaliyBondarenko1982, line 27: make constant for initial newMovie state outside of component and use it in two places (useState and resetForm)

VitaliyBondarenko1982 (changes requested):
Good job.
Needs some improvements.
Check such behavior.
Now I can fill form with spaces and add new movie with such value.
Screenshot from 2023-08-20 16-32-54
Screenshot from 2023-08-20 16-32-34

VitaliyBondarenko1982 (approved):
Well done.

PR #1746, opened 2023-08-22
done

maxonshev:
https://maxonshev.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -7,15 +7,17 @@ type Props = {
 };
 
 export const NewMovie: React.FC<Props> = ({ onAdd }) => {
-  const [count, setCount] = useState(0);
-
-  const [newMovie, setNewMovie] = useState({
+  const EMPTY_MOVIE = {
     title: '',
     description: '',
     imgUrl: '',
     imdbUrl: '',
     imdbId: '',
-  });
+  };
+
+  const [count, setCount] = useState(0);
+
+  const [newMovie, setNewMovie] = useState(EMPTY_MOVIE);
 
   const handleChange = (name: string, value: string) => {
     setNewMovie({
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,101 @@
...
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,101 @@
...
+        onChange={newTitle => setTitle(newTitle)}
         required
       />
 
       <TextField
         name="description"
         label="Description"
-        value=""
+        value={description}
+        onChange={newDescription => setDescription(newDescription)}
```
etojeDenys, line 70:
```suggestion
        onChange={setDescription}
```

etojeDenys (changes requested):
i can just write spaces in the title field and add a new movie, let's fix this case with `trim`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,102 @@
...
+
+  const reset = () => {
+    setNewMovie({
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
+  };
```
IvanFesenko, line 35: To avoid code duplicating you can create varaible with empty object and use it 
```
const EMPTY_MOVIE = {
    title: '',
    description: '',
    imgUrl: '',
    imdbUrl: '',
    imdbId: ''
};
```
```suggestion
  const [newMovie, setNewMovie] = useState(EMPTY_MOVIE );

  const handleChange = (name: string, value: string) => {
    setNewMovie({
      ...newMovie,
      [name]: value,
    });
  };

  const reset = () => {
    setNewMovie(EMPTY_MOVIE);
  };
```

IvanFesenko (approved):
Well done

PR #1751, opened 2023-08-27
add form

Orchibald:
[DEMO LINK](https://orchibald.github.io/react_movies-list-add-form/)


etojeDenys (changes requested):
I can create the film only with spaces
image
let's fix that

VitaliyBondarenko1982 (approved):
Looks good

PR #1752, opened 2023-08-27
Add movie form

AmadeuszAndroid:
[DEMO LINK](https://amadeuszandroid.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,32 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+
+const movies = moviesFromServer.map((movie) => ({
+  ...movie,
+  movieId: movie.imdbId,
+}));
 
 export const App = () => {
+  const [moviesCount, increaseMoviesCount] = useState(movies.length);
+
+  const onAdd = (movie: Movie) => {
+    if (moviesCount === movies.length) {
+      movies.push(movie);
+      increaseMoviesCount(oldMoviesCount => oldMoviesCount + 1);
+    }
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={onAdd} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,50 +1,96 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+const emptyString = '';
+
+type NewMovieProps = {
+  onAdd: (movie: Movie) => void
+};
+
+export const NewMovie = ({ onAdd }: NewMovieProps) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, increaseCount] = useState(3);
+  const [title, setTitle] = useState(emptyString);
+  const [description, setDescription] = useState(emptyString);
+  const [imgUrl, setImgUrl] = useState(emptyString);
+  const [imdbUrl, setImdbUrl] = useState(emptyString);
+  const [imdbId, setImdbId] = useState(emptyString);
+  const addMovie = (event: React.FormEvent) => {
+    event.preventDefault();
+    onAdd({
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+      movieId: `${Date.now()}${imdbId}`,
+    });
+    increaseCount(oldCount => oldCount + 1);
+    setTitle(emptyString);
+    setDescription(emptyString);
+    setImgUrl(emptyString);
+    setImdbUrl(emptyString);
+    setImdbId(emptyString);
+  };
 
   return (
-    <form className="NewMovie" key={count}>
+    <form
+      className="NewMovie"
+      key={count}
+      onSubmit={(event) => addMovie(event)}
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={newTitle => setTitle(newTitle)}
         required
       />
 
       <TextField
         name="description"
         label="Description"
-        value=""
+        value={description}
+        onChange={newDescription => setDescription(newDescription)}
       />
 
       <TextField
         name="imgUrl"
         label="Image URL"
-        value=""
+        value={imgUrl}
+        onChange={newUrl => setImgUrl(newUrl)}
+        required
       />
 
       <TextField
         name="imdbUrl"
         label="Imdb URL"
-        value=""
+        value={imdbUrl}
+        onChange={newUrl => setImdbUrl(newUrl)}
+        required
       />
 
       <TextField
         name="imdbId"
         label="Imdb ID"
-        value=""
+        value={imdbId}
+        onChange={newId => setImdbId(newId)}
+        required
       />
 
       <div className="field is-grouped">
         <div className="control">
           <button
+            disabled={
+              title === emptyString
+              || imgUrl === emptyString
+              || imdbUrl === emptyString
+              || imdbId === emptyString
+            }
             type="submit"
             data-cy="submit-button"
             className="button is-link"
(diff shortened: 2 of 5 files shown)


src/App.tsx
```diff
@@ -1,16 +1,32 @@
...
+  ...movie,
+  movieId: movie.imdbId,
+}));
 
 export const App = () => {
+  const [moviesCount, increaseMoviesCount] = useState(movies.length);
+
+  const onAdd = (movie: Movie) => {
+    if (moviesCount === movies.length) {
+      movies.push(movie);
```
darokrk, line 18: Directly manipulating state in React is not a good practice. Instead of pushing movie into movies, use a setState function to create a new array with the added movie. Also function name is not much descriptible, try use something like this:

```
const addNewMovie = (newMovie: Movie) => {
    setMoviesList(initialMovies => {
      return [...initialMovies,
        newMovie];
    });
  };
```

src/App.tsx
```diff
@@ -1,16 +1,32 @@
...
+const movies = moviesFromServer.map((movie) => ({
+  ...movie,
+  movieId: movie.imdbId,
+}));
 
 export const App = () => {
+  const [moviesCount, increaseMoviesCount] = useState(movies.length);
+
+  const onAdd = (movie: Movie) => {
+    if (moviesCount === movies.length) {
```
darokrk, line 17: The if check here is not needed. If a new movie is to be added, it can directly be added to the movies state without this check.

src/App.tsx
```diff
@@ -1,16 +1,32 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+
+const movies = moviesFromServer.map((movie) => ({
```
darokrk, line 8: You are directly manipulating the 'movies' variable that is outside the component. This is not a recommended practice in React. Use a state variable instead for storing movies.

`const [moviesList, setMoviesList] = useState(moviesFromServer);`

src/App.tsx
```diff
@@ -1,16 +1,32 @@
...
+  ...movie,
+  movieId: movie.imdbId,
+}));
 
 export const App = () => {
+  const [moviesCount, increaseMoviesCount] = useState(movies.length);
```
darokrk, line 14: The 'moviesCount' state variable seems redundant. The number of movies can always be derived from the movies state itself by accessing its length property.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,96 @@
...
+
+type NewMovieProps = {
+  onAdd: (movie: Movie) => void
+};
+
+export const NewMovie = ({ onAdd }: NewMovieProps) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, increaseCount] = useState(3);
```
darokrk, line 14: The initial value for count state is set to 3 which doesn't seem to have a clear purpose. It might be better to start it at 0.

src/App.tsx
```diff
@@ -1,16 +1,32 @@
...
+      movies.push(movie);
+      increaseMoviesCount(oldMoviesCount => oldMoviesCount + 1);
+    }
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
```
darokrk, line 26: `<MoviesList movies={moviesList} />`

src/App.tsx
```diff
@@ -1,16 +1,32 @@
...
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={onAdd} />
```
darokrk, line 29: `<NewMovie onAdd={addNewMovie} />`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,96 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+const emptyString = '';
```
darokrk, line 5: remove this variable, is redundant, for better readability just use explicitly an empty string in state:

```
  const [imgUrl, setImgUrl] = useState('');
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,96 @@
...
+export const NewMovie = ({ onAdd }: NewMovieProps) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, increaseCount] = useState(3);
+  const [title, setTitle] = useState(emptyString);
+  const [description, setDescription] = useState(emptyString);
+  const [imgUrl, setImgUrl] = useState(emptyString);
+  const [imdbUrl, setImdbUrl] = useState(emptyString);
+  const [imdbId, setImdbId] = useState(emptyString);
```
darokrk, line 19: add space between state and functions for better readability

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,96 @@
...
+  const addMovie = (event: React.FormEvent) => {
```
darokrk, line 20: this function should be renamed to `handleSubmit` rather then `addMovie`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,96 @@
...
+    event.preventDefault();
+    onAdd({
```
darokrk, line 22: you can create this object into variable to improve readability, e.g.:

```
const newMovie: Movie = {
      title,
      description,
      imgUrl,
      imdbUrl,
      imdbId,
      movieId: `${Date.now()}${imdbId}`
    };

    onAdd(newMovie);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,96 @@
...
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+      movieId: `${Date.now()}${imdbId}`,
+    });
+    increaseCount(oldCount => oldCount + 1);
```
darokrk, line 30: this reset state can be also extracted to separate function to improve readability, e.g:

```
  const resetFormValues = () => {
    setTitle('');
    setDescription('');
    setImgUrl('');
    setImdbUrl('');
    setImdbId('');
  };
  
  resetFormValues();
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,96 @@
...
+        value={imdbId}
+        onChange={newId => setImdbId(newId)}
+        required
       />
 
       <div className="field is-grouped">
         <div className="control">
           <button
+            disabled={
+              title === emptyString
```
darokrk, line 89: this check can be also extracted to variable called something like `const isButtonEnabled = .....` and the used here  :
```
<button
            disabled={isButtonEnabled}
>
```

darokrk (changes requested):
Not bad 👏🏽  added suggestion for code improvements 🛠️  👨🏽‍💻

src/App.tsx
AmadeuszAndroid: I added I renamed function and added movies to state.

src/App.tsx
AmadeuszAndroid: I deleted this check.

src/App.tsx
AmadeuszAndroid: I added movies to state.

src/App.tsx
AmadeuszAndroid: I deleted `moviesCount` from state.

src/components/NewMovie/NewMovie.tsx
AmadeuszAndroid: I set value to 0.

src/App.tsx
AmadeuszAndroid: I changed to `moviesList`.

src/App.tsx
AmadeuszAndroid: I renamed function to `addNewMovie`.

src/components/NewMovie/NewMovie.tsx
AmadeuszAndroid: I removed variable.

src/components/NewMovie/NewMovie.tsx
AmadeuszAndroid: I renamed function.

src/components/NewMovie/NewMovie.tsx
AmadeuszAndroid: I added new line between states and functions.

src/components/NewMovie/NewMovie.tsx
AmadeuszAndroid: I created variable.

src/components/NewMovie/NewMovie.tsx
AmadeuszAndroid: I extracted function.

src/components/NewMovie/NewMovie.tsx
AmadeuszAndroid: I created function.

mykhalenych (approved):

PR #1754, opened 2023-08-30
added autofill for imdbId

domkab:
https://domkab.github.io/react_movies-list-add-form


lerastarynets (changes requested):
I can now add film with all properties set to white space only, pls fix
Screenshot at Aug 30 17-35-47

anastasiia-tilikina (changes requested):
Well done!
let's fix failed tests:
I should be able to add a movie without a description.
`NewMovie` prop should be called `onAdd`, I belive tests require it.

Also, let's move validation logic from `TextField` to `NewMovie`

lerastarynets (approved):
You're doing a fantastic job, keep going!

PR #1763, opened 2023-09-04
Solution

oksana-vas:
[DEMO LINK](https://oksana-vas.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -2,93 +2,92 @@ import React, { useState } from 'react';
 import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
+const initialMovieState = {
+  title: '',
+  description: '',
+  imgUrl: '',
+  imdbUrl: '',
+  imdbId: '',
+};
+
 type Props = {
   onAdd: (movie: Movie) => void;
 };
 
 export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   const [count, setCount] = useState(0);
-  const [title, setTitle] = useState('');
-  const [description, setDescription] = useState('');
-  const [imgUrl, setImgUrl] = useState('');
-  const [imdbUrl, setImdbUrl] = useState('');
-  const [imdbId, setImdbId] = useState('');
+  const [newMovie, setNewMovie] = useState<Movie>(initialMovieState);
 
-  const reset = (event: React.FormEvent) => {
-    event.preventDefault();
-
-    setTitle('');
-    setDescription('');
-    setImgUrl('');
-    setImdbUrl('');
-    setImdbId('');
+  const reset = () => {
+    setNewMovie(initialMovieState);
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const reset = (event: React.FormEvent) => {
+    event.preventDefault();
```
Viktor-Kost, line 18: `event.preventDefault()` in the `reset` function is not necessary

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```

1. create a variable outside the Component
```
const initialMovieState = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: ''
}
```

2. use one state for movie fields
3. pass `initialMovieState` object when you invoke reset `function`
4. whenever you need to change value of any field you should pass two arguments (key, value) and then just change a value of a movie object by key

```suggestion
  const [count, setCount] = useState(0);
  const [movie, setMovie] = useState(initialMovieState);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+
+    onAdd({
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    });
+
+    setCount(newCount => newCount + 1);
```
Viktor-Kost, line 38:
```suggestion
    setCount(prevCount => prevCount + 1);
```

src/components/TextField/TextField.tsx
```diff
@@ -22,14 +22,14 @@ export const TextField: React.FC<Props> = ({
   label = name,
   placeholder = `Enter ${label}`,
   required = false,
-  onChange = () => {},
+  onChange = () => { },
```
Viktor-Kost, line 25:
```suggestion
  onChange = () => {},
```

Viktor-Kost (changes requested):

IvanFesenko (approved):
Well done

PR #1764, opened 2023-09-04
Develop

anpiet:
[DEMO LINK](https://anpiet.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,56 +1,129 @@
...
+      imdbId,
+    })
+  );
 
   return (
-    <form className="NewMovie" key={count}>
+    <form
+      className="NewMovie"
+      key={count}
+      onSubmit={(event) => {
```
HaniaNassalska, line 38: Maybe you should write it in a function above.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,56 +1,129 @@
...
-        value=""
+        value={imdbId}
+        required
+        onChange={(event) => {
+          setImdbId(event);
+        }}
       />
 
       <div className="field is-grouped">
         <div className="control">
```
HaniaNassalska, line 101: as above :)

HaniaNassalska (approved):
Everything is clear and working 👍

src/components/NewMovie/NewMovie.tsx
HaniaNassalska: this is the part i'm talking about
`(title.length === 0
          || imgUrl.length === 0
          || imdbUrl.length === 0
          || imdbId.length === 0)
`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,56 +1,129 @@
...
+      onSubmit={(event) => {
+        event.preventDefault();
+        setCount(count + 1);
+        onAdd(movie);
+        setTitle('');
+        setDescription('');
+        setImgUrl('');
+        setImdbUrl('');
+        setImdbId('');
+      }}
```
choeqq, line 47: it would be good idea to have separate function for that

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,56 +1,129 @@
...
-            type="submit"
-            data-cy="submit-button"
-            className="button is-link"
-          >
-            Add
-          </button>
+          {(title.length === 0
+          || imgUrl.length === 0
+          || imdbUrl.length === 0
+          || imdbId.length === 0)
```
choeqq, line 105: this condition can also be extracted into variable - for simplicity

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,56 +1,129 @@
...
+            ? (
+              <button
+                type="submit"
+                data-cy="submit-button"
+                className="button is-link"
+                disabled
+              >
```
choeqq, line 112: instead of rendering 2 buttons you can set disabled conditionally
```suggestion
              <button
                type="submit"
                data-cy="submit-button"
                className="button is-link"
                disabled={someBooleanValue}
              >
```

choeqq (changes requested):

PR #1770, opened 2023-09-06
solution

drhtka:
https://drhtka.github.io/react_movies-list-add-form/


src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+  const movie = (newMovie: Movie) => {
```
DarkMistyRoom, line 11:
```suggestion
  const addMovie = (newMovie: Movie) => {
```

DarkMistyRoom: use verbs for functions' names

DarkMistyRoom (changes requested):
use `trim` to avoid the possibility to add only white spaces
image

mykhalenych (approved):

PR #1771, opened 2023-09-06
Develop

Wesses:
[DEMO LINK](https://Wesses.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,92 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (newMovie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const disableSubmit = !title || !imgUrl || !imdbUrl || !imdbId;
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,92 @@
...
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const disableSubmit = !title || !imgUrl || !imdbUrl || !imdbId;
```
davinnchii, line 20: I think better name here is `isSubmitDisabled` or `isButtonDisabled`

davinnchii (approved):
GJ

Moroz-Dmytro (changes requested):
image

I am able to add movies even when I have invalid URLs and see an error message. Also, maybe would be good to give more descriptive error message(because now my input is not empty but I see error, we should inform user that URL is invalid)

roman-mirzoian (changes requested):
Ok, but now I can add a movie without a title (if I put empty spaces in there)
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,102 @@
...
+  description: '',
+  imgUrl: '',
+  imdbUrl: '',
+  imdbId: '',
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+
+  const [movieState, setMovieState] = useState<Movie>(initialState);
```
Moroz-Dmytro, line 21:
```suggestion
  const [newMovie, setNewMovie] = useState<Movie>(initialState);
```

src/components/TextField/TextField.tsx
```diff
@@ -55,6 +60,10 @@ export const TextField: React.FC<Props> = ({
       {hasError && (
         <p className="help is-danger">{`${label} is required`}</p>
       )}
+
+      {!hasError && hasValidationError && (
```
Moroz-Dmytro, line 64: Create variable for this

Moroz-Dmytro (changes requested):
Would be good to provide message why I can't click "Add" movie, because now it's confusing for user what is wrong

image

roman-mirzoian (approved):
Well done 👍

PR #1788, opened 2023-09-06
Solution

reseneweb:
https://reseneweb.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,129 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const requiredCompleted = title && imgUrl && imdbUrl && imdbId;
+
+  const reset = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+    setCount(0);
+  };
+
+  const handleAdd = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!requiredCompleted) {
+      return;
+    }
+
+    onAdd({
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    });
+
+    reset();
+    setCount(1);
+  };
+
+  const handleTitleChange
+  = (newValue: string) => {
+    setTitle(newValue);
+  };
+
+  const handleDescriptionChange
+  = (newValue: string) => {
+    setDescription(newValue);
+  };
+
+  const handleImgUrlChange
+  = (newValue: string) => {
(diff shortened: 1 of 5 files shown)


plodnikate (changes requested):
Hi, in general looks good, but why you need to duplicate the entered value next to label ? 
Screenshot 2023-09-06 at 20 15 25

plodnikate (changes requested):
Please, fix this moment.

maxim2310 (changes requested):
Good job! Let's improve your code a bit and add validation for URL its not complicated

image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,140 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```

1. create a variable outside the Component
```
const initialMovieState = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: ''
}
```

2. use one state for movie fields
3. pass `initialMovieState` object when you invoke reset `function`
4. whenever you need to change value of any field you should pass two arguments (key, value) and then just change a value of a movie object by key. In this case you shouldn't have different handlers for your fields

Example:
```
const handleInputChange = (key, value) = {
  setMovie(prevFields => {
    return {
      ...prevFields,
      [key]: value
    }
  })
}
```


```suggestion
  const [movie, setMovie] = useState(initialMovieState);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,140 @@
...
+    setLinkImgValidity(false);
+  };
+
+  const handleAdd = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!requiredCompleted) {
+      return;
+    }
+
```
Viktor-Kost, line 51: before adding a movie you should validate fields.
Also you should avoid adding values just with spaces (you can use `trim`)

Screenshot 2023-09-07 at 20 03 59

Viktor-Kost (changes requested):

Viktor-Kost (changes requested):
.

src/components/NewMovie/NewMovie.tsx
mykhalenych: https://github.com/mate-academy/react_movies-list-add-form/pull/1788/files#r1318897323
didn't fix it

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,140 @@
...
+  const requiredCompleted
+  = title && imgUrl && imdbUrl && imdbId && linkImdbValidity && linkImgValidity;
+
+  const pattern = new RegExp(
+    '^((([A-Za-z]{3,9}:(?://)?)|(?:www\\.|[-;:&=+$,\\w]+@))[A-Za-z0-9.-]+'
+    + '((?:\\/[+~%\\/.\\w-_]*)?\\??(?:[-+=&;%@,.\\w_]*)#?('
+    + '?:[,.!/\\\\\\w]*))?)$',
+  );
+
+  const validationOfImgLink = (inputText: string) => {
```
mykhalenych, line 27: you links validations aren't working

https://github.com/mate-academy/react_movies-list-add-form/assets/53228259/ebaa82b0-787d-402a-aad7-996f9a191ede

src/components/NewMovie/NewMovie.tsx
mykhalenych: https://github.com/mate-academy/react_movies-list-add-form/pull/1788/files#r1318899412
didn't fix it

mykhalenych (changes requested):

maxim2310 (approved):
Good work!!!

PR #1792, opened 2023-09-06
add functional for movies form

rostyslav48:
[DEMO LINK](https://rostyslav48.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,7 +19,7 @@ export const App = () => {
         <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie onAdd={(movie: Movie) => addMovie(movie)} />
+        <NewMovie onAdd={addMovie} />
       </div>
     </div>
   );
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,25 @@
+import { useState } from 'react';
+
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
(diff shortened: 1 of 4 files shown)


mr-olezha (approved):
Looks good 🔥

src/App.tsx
```diff
@@ -1,16 +1,25 @@
...
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie: Movie) => addMovie(movie)} />
```
Moroz-Dmytro, line 22:
```suggestion
        <NewMovie onAdd={addMovie} />
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
Moroz-Dmytro, line 17: Let's refactor your solution and use only 1 state for all form fields
It will be next object`{ title: '', description: '', ...}`

Moroz-Dmytro (changes requested):
Now I'm able to add movie with empty fields, let's fix this 😉 

image

src/components/NewMovie/NewMovie.tsx
rostyslav48: I have reduced number of `states`. However Anna said that it is not a good practice to do 1 state to all fields, is it?

roman-mirzoian (approved):
Well done 👍

PR #1793, opened 2023-09-06
added logic to add new movie and pattern for url

AntonHerasymov:
[DEMO LINK](https://AntonHerasymov.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  const addMovie = (newMovie: Movie) => {
+    setMovies((currentMovies : Movie[]) => [...currentMovies, newMovie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(addMovie)} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,90 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const formReset = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbId('');
+    setImdbId('');
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    onAdd({
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    });
+
+    formReset();
+
+    setCount((prevCount) => prevCount + 1);
+  };
 
   return (
-    <form className="NewMovie" key={count}>
+    <form
+      className="NewMovie"
+      key={count}
+      onSubmit={handleSubmit}
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(newValue) => setTitle(newValue)}
         required
       />
 
       <TextField
         name="description"
         label="Description"
-        value=""
+        value={description}
+        onChange={(newValue) => setDescription(newValue)}
       />
 
       <TextField
         name="imgUrl"
         label="Image URL"
-        value=""
+        value={imgUrl}
+        onChange={(newValue) => setImgUrl(newValue)}
+        required
       />
 
       <TextField
         name="imdbUrl"
         label="Imdb URL"
-        value=""
+        value={imdbUrl}
+        onChange={(newValue) => setImdbUrl(newValue)}
+        required
       />
 
       <TextField
         name="imdbId"
         label="Imdb ID"
-        value=""
+        value={imdbId}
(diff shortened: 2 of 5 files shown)


src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  const addMovie = (newMovie: Movie) => {
```
dffUqp, line 11: Component events start with the prefix `on` but the function that handles these events should not have this prefix.

Let's change it

src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
+    setMovies((currentMovies : Movie[]) => [...currentMovies, newMovie]);
```
dffUqp, line 12:
```suggestion
    setMovies((currentMovies) => [...currentMovies, newMovie]);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,90 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
dffUqp, line 17: Can be replaced with one useState, using an object like in the example below.
```javascript
const [formData, setFormData] = useState({
  title: "",
  description: "",
  imgUrl: "",
  ...
});
```
Create const outside your component where you can save your `initialFormData` which will have all initial values
```jsx
const initialFormData = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: '',
};
```


To change this `formData` you can create a handler like this
```jsx
 const handleDataChange = (key: string, value: string) => {
    setFormData((currentState) => ({
      ...currentState,
      [key]: value,
    }));
  };
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +93,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!title || !imgUrl || !imdbUrl || !imdbId}
```
dffUqp, line 96: Let's create separate variable for this

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,90 @@
...
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(newValue) => setTitle(newValue)}
```
dffUqp, line 55: Just keep in mind that a components like `TextField` call your function and provide all the props, so we don't need to create an extra function.
```suggestion 
        onChange={setTitle}
```

src/components/TextField/TextField.tsx
```diff
@@ -31,6 +32,10 @@ export const TextField: React.FC<Props> = ({
   const [touched, setTouched] = useState(false);
   const hasError = touched && required && !value;
 
+  const isValidUrl = (value && (label === 'Image URL' || label === 'Imdb URL'))
+    ? urlPattern.test(value)
+    : true;
```
dffUqp, line 37: `TextField` component should not contain any business logic.

If I decided to create a new form with a hundred `TextField` and none of these would be URL, I think this logic would be unnecessary.

So I propose to create new prop like `validationPattern`, `isValid`, or just `error` and handle your logic there.

```jsx
      <TextField
        name="imdbUrl"
        label="Imdb URL"
        value={imdbUrl}
        onChange={(newValue) => setImdbUrl(newValue)}
        // error={urlPattern.test(value)}
        // validationPattern={urlPattern}
        // isValid={urlPattern.test(value)}
        required
      />
 ```
 
 (Also you can add `helperText` or `errorMessage` props if needed)

File: src/patterns/pattern.ts [hunk 7]
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars, max-len
+export const urlPattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
```
dffUqp, line 2: You have a `patterns` folder and I think you should have `urlPattern.ts`, etc. in that folder, not `pattern.ts`.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,90 @@
...
+
+    setCount((prevCount) => prevCount + 1);
+  };
 
   return (
-    <form className="NewMovie" key={count}>
+    <form
+      className="NewMovie"
+      key={count}
+      onSubmit={handleSubmit}
```
dffUqp, line 47: I can submit a form that contains only spaces.

image

dffUqp (changes requested):
LGTM:fire: but needs a few fixes:upside_down_face:

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,104 @@
...
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [formData, setFormData] = useState(initialFormData);
+
+  const formReset = () => {
```
Moroz-Dmytro, line 24: function name should start with a verb

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +107,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!isAllFieldsValid()}
```
Moroz-Dmytro, line 110: Better not to use function inside JSX, create variable for this

Moroz-Dmytro (changes requested):
I'm not able to create movie even when all required fields are inputted

image

I can see 2 error messages at the same time

image

I can't create movie, but I don't see any messages why

image

src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(handleAddMovie)} />
```
roman-mirzoian, line 21:
```suggestion
        <NewMovie onAdd={handleAddMovie} />
```

roman-mirzoian (approved):
Well done 👍

PR #1807, opened 2023-09-07
feat: make possible to add a new movie

FrontKid:
[DEMO LINK](https://Frontkid.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -7,71 +7,33 @@ type TNewMovie = {
 };
 
 export const NewMovie: FC<TNewMovie> = ({ onAdd }) => {
+  const initialInputFields = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
+
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
   const [count, setCount] = useState(0);
-  const [inputFields, setInputFields] = useState({
-    title: '',
-    description: '',
-    imageUrl: '',
-    imdbUrl: '',
-    imdbId: '',
-  });
+  const [inputFields, setInputFields] = useState(initialInputFields);
 
   const handleFormSubmit = (event: React.FormEvent<HTMLFormElement>) => {
     event.preventDefault();
 
-    onAdd({
-      title: inputFields.title,
-      description: inputFields.description,
-      imgUrl: inputFields.imageUrl,
-      imdbUrl: inputFields.imdbUrl,
-      imdbId: inputFields.imdbId,
-    });
+    onAdd(inputFields);
 
-    setInputFields({
-      title: '',
-      description: '',
-      imageUrl: '',
-      imdbUrl: '',
-      imdbId: '',
-    });
+    setInputFields(initialInputFields);
 
     setCount(prev => prev + 1);
   };
 
-  const handleTitleChange = (currentValue: string) => {
+  const handleChange = (currentValue: string, name:string) => {
     setInputFields((prevState) => ({
       ...prevState,
-      title: currentValue,
-    }));
-  };
-
-  const handleDescriptionChange = (currentValue: string) => {
-    setInputFields((prevState) => ({
-      ...prevState,
-      description: currentValue,
-    }));
-  };
-
-  const handleImageUrlChange = (currentValue: string) => {
-    setInputFields((prevState) => ({
-      ...prevState,
-      imageUrl: currentValue,
-    }));
-  };
-
-  const handleImdbUrlChange = (currentValue: string) => {
-    setInputFields((prevState) => ({
-      ...prevState,
-      imdbUrl: currentValue,
-    }));
-  };
-
-  const handleImdbIdChange = (currentValue: string) => {
-    setInputFields((prevState) => ({
-      ...prevState,
-      imdbId: currentValue,
+      [name]: currentValue,
     }));
   };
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,16 +1,31 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  const onAdd = (newMovie: Movie) => {
```
roman-mirzoian, line 11:
```suggestion
  const handleMovieAdd = (newMovie: Movie) => {
```

src/App.tsx
```diff
@@ -1,16 +1,31 @@
...
+    setMovies((prev) => {
+      return [
+        ...prev,
+        newMovie,
+      ];
+    });
```
roman-mirzoian, line 17:
```suggestion
    setMovies((prev) => [...prev, newMovie]);
```

src/App.tsx
```diff
@@ -1,16 +1,31 @@
...
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie
+          onAdd={onAdd}
+        />
```
roman-mirzoian, line 28:
```suggestion
        <NewMovie onAdd={onAdd} />
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,130 @@
...
+      description: '',
+      imageUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
+
+    setCount(prev => prev + 1);
+  };
+
+  const onTitleHandler = (currentValue: string) => {
```
roman-mirzoian, line 43:
```suggestion
  const handleTitleChange = (currentValue: string) => {
```
Fix naming for other handlers as well.

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.
Now you can add a movie with empty information by putting empty spaces in each input:
https://gyazo.com/1e46cc4e36099e2b7fbe36b596d76d4e

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,130 @@
...
+      imdbId: inputFields.imdbId,
+    });
+
+    setInputFields({
+      title: '',
+      description: '',
+      imageUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
```
BudnikOleksii, line 38: You can create an object with default values and use it when initialize and here when reset fields

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,130 @@
...
+  const handleFormSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    onAdd({
+      title: inputFields.title,
+      description: inputFields.description,
+      imgUrl: inputFields.imageUrl,
+      imdbUrl: inputFields.imdbUrl,
+      imdbId: inputFields.imdbId,
+    });
```
BudnikOleksii, line 30: You don't need to repeat all fields, you can just pass the `inputFields` object

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,130 @@
...
+      imdbUrl: currentValue,
+    }));
+  };
+
+  const handleImdbIdChange = (currentValue: string) => {
+    setInputFields((prevState) => ({
+      ...prevState,
+      imdbId: currentValue,
+    }));
+  };
```
BudnikOleksii, line 76: All handlers look similar, you can create one handler and use it

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,130 @@
...
+
+  const isAllFieldValid = Object
+    .entries(inputFields)
+    .filter(([key]) => key !== 'description')
+    .every(([, val]) => val);
```
BudnikOleksii, line 81: You can just check values, you don't need keys here, something like this:
```suggestion
  const isAllFieldValid = Object
    .values(inputFields)
    .every((value) => value);
```

BudnikOleksii (changes requested):
Good work 👍 
But could be even better, check comments

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,92 @@
...
+};
+
+export const NewMovie: FC<TNewMovie> = ({ onAdd }) => {
+  const initialInputFields = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
```
Moroz-Dmytro, line 16: Better to move this constant to separate file and name in UPPER_CASE

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,92 @@
...
+    event.preventDefault();
+
+    onAdd(inputFields);
+
+    setInputFields(initialInputFields);
+
+    setCount(prev => prev + 1);
+  };
+
+  const handleChange = (currentValue: string, name:string) => {
```
Moroz-Dmytro, line 33:
```suggestion
  const handleInputChange = (currentValue: string, name:string) => {
```

Moroz-Dmytro (approved):

PR #1809, opened 2023-09-07
add movie list add solution

PavloMykhalov:


anastasiiavorobiova (changes requested):
Great work! Consider implementing trimming before submit Screenshot 2023-09-07 at 17 40 06

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review. 
It is necessary to disable the button when there is no data in the fields, so as not to add empty data
image
image

volodymyr-soltys97 (approved):
Great work 🔥

PR #1813, opened 2023-09-07
add solution

dimarogkov:
[DEMO LINK](https://dimarogkov.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,93 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (newMovie: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [newMovie, setNewMovie] = useState<Movie>({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,94 @@
...
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):
let's show the error when the user enters only spaces in field

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+    onAdd(newMovie);
+    resetForm();
+    setCount(prev => prev + 1);
+  };
 
   return (
-    <form className="NewMovie" key={count}>
+    <form
+      className="NewMovie"
+      key={count}
```
DarkMistyRoom, line 51: no need to set `key` here. Use it only for the lists

DarkMistyRoom (changes requested):
please show the errors when inputs are filled with white spaces only. If you need help with it, please ask in the chat

maxim2310 (changes requested):
please show the errors when inputs are filled with white spaces only. If you need help with it, please ask in the chat

etojeDenys (changes requested):
still not fixed
image

dimarogkov:
Name at least one reason why I should take away what should be on condition?

dimarogkov:
image

VitaliyBondarenko1982 (approved):
GJ!

PR #1820, opened 2023-09-10
add task solution

cty2802:
[DEMO LINK](https://cty2802.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,22 @@
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
 
 export const App = () => {
+  const [preparedMovies, setPreparedMovies] = useState(moviesFromServer);
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={preparedMovies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie) => {
+          setPreparedMovies([...preparedMovies, movie]);
+        }}
+        />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,105 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd?: (movie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd = () => {} }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
   const [count] = useState(0);
 
+  const [title, setTtitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const reset = () => {
+    setTtitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
+
+  function isDisabled(): boolean {
+    return !(title && imgUrl && imdbUrl && imdbId);
+  }
+
+  const handleSubmit = (event: React.FormEvent<Element>) => {
+    event.preventDefault();
+
+    const newMovie: Movie = {
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    };
+
+    onAdd(newMovie);
+
+    reset();
+  };
+
   return (
-    <form className="NewMovie" key={count}>
+    <form
+      className="NewMovie"
+      key={count}
+      onSubmit={handleSubmit}
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(event) => {
+          setTtitle(event);
+        }}
         required
       />
 
       <TextField
         name="description"
         label="Description"
-        value=""
+        value={description}
+        onChange={(event) => {
+          setDescription(event);
+        }}
       />
 
       <TextField
         name="imgUrl"
         label="Image URL"
-        value=""
+        value={imgUrl}
+        onChange={(event) => {
+          setImgUrl(event);
+        }}
+        required
       />
 
       <TextField
         name="imdbUrl"
         label="Imdb URL"
-        value=""
+        value={imdbUrl}
+        onChange={(event) => {
+          setImdbUrl(event);
+        }}
+        required
       />
 
       <TextField
         name="imdbId"
         label="Imdb ID"
-        value=""
+        value={imdbId}
+        onChange={(event) => {
+          setImdbId(event);
+        }}
+        required
       />
 
       <div className="field is-grouped">
@@ -48,6 +108,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={isDisabled()}
           >
             Add
           </button>


src/App.tsx
```diff
@@ -1,16 +1,22 @@
...
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={preparedMovies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie) => {
+          setPreparedMovies([...preparedMovies, movie]);
+        }}
```
Moroz-Dmytro, line 18: Create separate function for this callback

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,105 @@
...
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd?: (movie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd = () => {} }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
```
Moroz-Dmytro, line 11: Don't leave comments in your code

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,105 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd?: (movie: Movie) => void,
```
Moroz-Dmytro, line 6: Why it is optional?

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,105 @@
...
   // to reset touched status of all the `Field`s
   const [count] = useState(0);
 
+  const [title, setTtitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const reset = () => {
```
Moroz-Dmytro, line 20:
```suggestion
  const resetFormFields = () => {
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,105 @@
...
+    setTtitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
+
+  function isDisabled(): boolean {
+    return !(title && imgUrl && imdbUrl && imdbId);
+  }
```
Moroz-Dmytro, line 30: Better to make this as a constant not as a funciton

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,105 @@
...
+    const newMovie: Movie = {
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    };
+
+    onAdd(newMovie);
+
```
Moroz-Dmytro, line 44: You need to increase your `count` here

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,105 @@
...
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(event) => {
+          setTtitle(event);
+        }}
```
Moroz-Dmytro, line 62:
```suggestion
        onChange={setTtitle}
```

Moroz-Dmytro: The same in other places

Moroz-Dmytro (changes requested):

src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [preparedMovies, setPreparedMovies] = useState(moviesFromServer);
+
+  const addPreparedMovie = (movie: Movie) => {
+    setPreparedMovies([...preparedMovies, movie]);
```
roman-mirzoian, line 12: In order to work with the actual state, you need to take it from the callback:
```suggestion
    setPreparedMovies((prevMovieList) => [...prevMovieList, movie]);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd = () => {} }) => {
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
roman-mirzoian, line 16: If we want to manage one entity, it is better to keep its data in one state.

roman-mirzoian (changes requested):
Now you can add a movie with empty fields:
https://gyazo.com/0570b6114f791e4397acc05d7b094331

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,104 @@
...
+
+  function createNewMovie() {
+    return {
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    };
+  }
```
BudnikOleksii, line 20: I don't get why `createNewMovie` returns an object with empty fields 🤷‍♂️ 

You can just create and use variable to set default values instead of function(also, create this variable outside of the component).

```javascript
const DEFAULT_MOVIE_DATA = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: '',
};
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,104 @@
...
+  };
+
+  const isDisabled = () => {
+    return !(
+      newMovie.title
+      && newMovie.imgUrl
+      && newMovie.imdbUrl
+      && newMovie.imdbId
+    );
+  };
```
BudnikOleksii, line 35:
```suggestion
  const isDisabled = !(
      newMovie.title
      && newMovie.imgUrl
      && newMovie.imdbUrl
      && newMovie.imdbId
    );
```
You don't need to do this check with a function, just use it as a variable

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,104 @@
...
       />
 
       <TextField
         name="imdbUrl"
         label="Imdb URL"
-        value=""
+        value={newMovie.imdbUrl}
+        onChange={(imdbUrl) => {
+          setNewMovie({ ...newMovie, imdbUrl });
+        }}
```
BudnikOleksii, line 90: Use functional setState when you need to update the state and use the previous state
```javascript
setNewMovie(prev => ({
  ...prev,
  imdbUrl,
}));
```
Also, your handlers look the same for all fields so let's create one handler for all of them

BudnikOleksii (changes requested):
Good work 👍 
But could be even better. Left some comments, check them

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,102 @@
...
       />
 
       <TextField
         name="imgUrl"
         label="Image URL"
-        value=""
+        value={newMovie.imgUrl}
+        onChange={(imgUrl) => {
+          setNewMovie({ ...newMovie, imgUrl });
+        }}
```
BudnikOleksii, line 76: You fixed only for the `imdbUrl`, other handlers still don't use functional setState.

Also, all you handlers have a one pattern, let's create one handler for all of them

BudnikOleksii (changes requested):
Good progress, but something needs to be fixed.
Your handlers look similar, so let's create one to handle all input fields changes.

And don't forget when you change state based on the previous value -> you need to use functional state change.

PR #1833, opened 2023-09-12
Solution

SerhiiVoitiuk:
[DEMO LINK](https://SerhiiVoitiuk.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,119 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, seDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setSmdbId] = useState('');
+  const [imgError, setImgError] = useState(false);
+  const [imdbError, setImdbError] = useState(false);
+
+  const reset = () => {
+    setTitle('');
+    seDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setSmdbId('');
+    setImgError(false);
+    setImdbError(false);
+  };
+
+  const urlRegex = /^(https?:\/\/)?(www\.)?[A-Za-z\d.-]+\.\w{2,6}\/?([^\s]*)?$/;
+
+  const handlSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!title || !imgUrl || !imdbUrl || !imdbId) {
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,119 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, seDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setSmdbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):
i can submit the form with spaces

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,120 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, seDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
IvanFesenko, line 15: Not fixed previous comment
2023-09-13_16h08_57

IvanFesenko (changes requested):
Some issue with validation. Only spaces - is not valid 
https://github.com/mate-academy/react_movies-list-add-form/assets/61904995/a4c8867d-6f0d-49f7-b443-2c7c341c888a

anastasiia-tilikina (changes requested):
not fixed

image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,108 @@
...
+
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [newMovie, setNewMovie] = useState<Movie>(initialInput);
+  const [imdbValidity, setImdbValidity] = useState(false);
+  const [imgValidity, setImgValidity] = useState(false);
```
etojeDenys, line 21: we don't need a state for these values, we can save it in some vars

etojeDenys (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,99 @@
...
+  const [newMovie, setNewMovie] = useState<Movie>(initialInput);
+
+  const pattern = /^(https?:\/\/)?(www\.)?[A-Za-z\d.-]+\.\w{2,6}\/?([^\s]*)?$/;
+
+  const handleChange = (field: string, value: string) => {
+    setNewMovie({ ...newMovie, [field]: value.trimStart() });
+  };
+
+  const isValidImdbUrl = newMovie.imdbUrl.match(pattern);
+  const isValidImgUrl = newMovie.imgUrl.match(pattern);
```
anastasiia-tilikina, line 28: use test method instead

anastasiia-tilikina (changes requested):
Use [test](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) for url validation

OleziO (approved):

PR #1846, opened 2023-09-13
solution

luka4ok:
Please, check my solution:
[DEMO LINK](https://luka4ok.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,28 @@
+import React, { useState } from 'react';
+
 import './App.scss';
+
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+
 import moviesFromServer from './api/movies.json';
 
-export const App = () => {
+import { Movie } from './types/Movie';
+
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = (newMovie: Movie) => {
+    setMovies([...movies, newMovie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addMovie} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,107 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (newMovie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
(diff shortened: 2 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,107 @@
...
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
Viktor-Kost, line 16: It's not advisable to generate such a large number of states when we only have one instance (movie). 

```
const initialMovieState = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: ''
}
```

Example: 

```
}
```


```suggestion
  const [movie, setMovie] = useState(initialMovieState);
```

src/App.tsx
```diff
@@ -1,16 +1,28 @@
...
 import moviesFromServer from './api/movies.json';
 
-export const App = () => {
+import { Movie } from './types/Movie';
+
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = (newMovie: Movie) => {
+    setMovies([...movies, newMovie]);
```
Viktor-Kost, line 16: use prev vale
```suggestion
    setMovies(prevMovies => [...prevMovies, newMovie]);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,107 @@
...
+
+  const isFormFilled = !title || !imgUrl || !imdbUrl || !imdbId;
```
Viktor-Kost, line 18: You validation should exclude spaces since now I can just press `Space` and set it as a value

src/App.tsx
VitaliyBodnar: agree, please do it

src/App.tsx
```diff
@@ -1,16 +1,28 @@
...
 import { NewMovie } from './components/NewMovie';
+
 import moviesFromServer from './api/movies.json';
 
-export const App = () => {
+import { Movie } from './types/Movie';
+
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+  // Adding new movie to MovieList
```
VitaliyBodnar, line 14: please remove all comments

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,89 @@
...
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    onAdd(movie);
+
+    setCount((currentCount) => currentCount + 1);
+
+    reset();
```
VitaliyBodnar, line 43: you function have a lot of spaces, it can be less
```suggestion
    event.preventDefault();

    onAdd(movie);
    setCount((currentCount) => currentCount + 1);
    reset();
```

VitaliyBodnar (changes requested):

maxim2310 (changes requested):

1. i can fill all field with spaces and create new card
2. also need to add validation for url

maxim2310 (changes requested):
↑↑↑↑↑

luka4ok:
> ↑↑↑↑↑

Due to a recent redeployment, filling with spaces is no longer possible.

Isn't validation optional? (https://github.com/mate-academy/react_movies-list-add-form#optional-advanced-validation)

Andriy-Shymkiv (approved):

PR #1848, opened 2023-09-13
done

Aleksey-Aleksey:
[DEMO LINK](https://Aleksey-Aleksey.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,90 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,90 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):

src/components/NewMovie/NewMovie.tsx
Aleksey-Aleksey: you know, your approach can be much better but for me its too comcplicated in terms of understanding, at all. Espesially handleChange function. ive just added a reset function) so, please let me go next)

anastasiia-tilikina: let's consider the suggested implementation though))
if you have any questions - feel free to ask them in chat))

anastasiia-tilikina (changes requested):
Check my comment

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,121 @@
...
+  // const [imdbUrl, setImdbUrl] = useState('');
+  // const [imdbId, setImdbId] = useState('');
+
+  // const reset = () => {
+  //   setTitle('');
+  //   setDescription('');
+  //   setImgUrl('');
+  //   setImdbUrl('');
+  //   setImdbId('');
+  // };
```
etojeDenys, line 50: remove all comments

etojeDenys (changes requested):

src/components/NewMovie/NewMovie.tsx
Aleksey-Aleksey: done

etojeDenys (approved):
good job

PR #1850, opened 2023-09-14
solution

MaciejKujawa:
[DEMO LINK](https://MaciejKujawa.github.io/react_movies-list-add-form/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ You have the `App` with the `MoviesList` and `NewMovie` form containing ready
 to use `TextField` components. Learn how it works and implement an ability to
 add movies from [IMDB](https://www.imdb.com/).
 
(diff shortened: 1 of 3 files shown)


AndrzejPec (changes requested):
I think the error message should be displayed when the user clicks on any field (except  'description'), leaves it empty, and then shifts focus away from the field by clicking elsewhere.

choeqq (changes requested):
Yes, I agree with @AndrzejPec ⬆️ 
Other than that, everything is fine

choeqq (changes requested):
Waiting for any updates @MaciejKujawa 👀

choeqq (approved):
Yup, your taak solution meeta the requirements - I should’ve approved it, but I thought the previous comment was valid, and I didn’t wanted you to miss it.
Good job ✅

PR #1852, opened 2023-09-14
Add task solution

NikitaOrliak:
[DEMO LINK](https://NikitaOrliak.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,90 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (moive: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const reset = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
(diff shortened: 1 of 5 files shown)


anastasiiavorobiova (changes requested):
Screenshot 2023-09-14 at 17 00 50
Great work! I was able to submit spaces only, consider fixing it

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,92 @@
...
+  onAdd: (moive: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
maxim2310, line 15: would be nice to use one object for this all state

maxim2310 (changes requested):
Good job! Let's improve your code a bit)

src/components/NewMovie/NewMovie.tsx
NikitaOrliak: I asked about this in Slack, but mentor said that I should use states here, because if I wanna create handleInput function for an object state, I need to take an "event: React.FormEvent<HTMLFormElement>" in this function, but in TextField.tsx I use event: string. 
Here's a link with my question: https://mate-parttime.slack.com/archives/C04GA609AQ6/p1694694046443569

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,94 @@
...
+
+const EMPTY = '';
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState(EMPTY);
+  const [description, setDescription] = useState(EMPTY);
+  const [imgUrl, setImgUrl] = useState(EMPTY);
+  const [imdbUrl, setImdbUrl] = useState(EMPTY);
+  const [imdbId, setImdbId] = useState(EMPTY);
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):

etojeDenys (approved):

PR #1854, opened 2023-09-15
sol

cardinal1991:
[DEMO LINK](https://cardinal1991.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
-export const App = () => {
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const onAddMovie = (newMovie: Movie) => {
+    setMovies([...movies, newMovie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAddMovie={onAddMovie} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,88 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
+type MovieProps = {
+  onAddMovie: (movie: Movie) => void
+};
+
+export const NewMovie: React.FC<MovieProps> = ({ onAddMovie }: MovieProps) => {
   const [count] = useState(0);
 
+  const [title, setTitle] = useState<string>('');
+  const [description, setDescription] = useState<string>('');
+  const [imgUrl, setImgUrl] = useState<string>('');
+  const [imdbUrl, setImdbUrl] = useState<string>('');
+  const [imdbId, setImdbId] = useState<string>('');
+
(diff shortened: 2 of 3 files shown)


Dolnys (approved):
Everything looks good and clear, well done!😎

choeqq (changes requested):
After submitting a form fields are cleared, which is great, but then your code checks if we can submit the form, and since fields are empty, your form validation yields an error, please fix that and you're good to go 🚀

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,88 @@
...
+      description,
+    };
+
+    onAdd(newMovie);
+
+    setTitle(' ');
+    setImgUrl(' ');
+    setImdbUrl(' ');
+    setImdbId(' ');
+    setDescription(' ');
```
pawel-peksa, line 38: Lack of no 5 from the problem statement:

> 5. Errors should not be shown after clearing the form (change its key to reinitialize the form);

Let's make use of `count` state to reinitialize the form

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +91,10 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={
+              // eslint-disable-next-line max-len
+              !title.trim() || !imgUrl.trim() || !imdbUrl.trim() || !imdbId.trim()
+            }
```
pawel-peksa, line 97:
```suggestion
            disabled={isValidForm() }
```
And then please move that logic to a function definition. This way we the code will be much cleaner and there should be no need for disabling eslint

pawel-peksa (changes requested):
Looks good, however there are couple of things that can be improved

src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const onAdd = (newMovie: Movie) => {
+    setMovies([...movies, newMovie]);
```
JoyCoffeeAddict, line 12: you should use `function syntax` when setting state based on the previous value. Read [here](https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state)

src/components/NewMovie/NewMovie.tsx
JoyCoffeeAddict: We've talked about `' '` spaces at q&a, but this point is better. @cardinal1991 try to use `count` and reset values to empty strings `''`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,94 @@
...
+    setImdbUrl('');
+    setImdbId('');
+    setDescription('');
+
+    setCount((prevCount) => prevCount + 1);
+  };
+
+  const isDisabled = () => {
+    return !title.trim() || !imgUrl.trim() || !imdbUrl.trim() || !imdbId.trim();
+  };
```
pawel-peksa, line 45: This logic here looks excactly the same as in `handleSubmit` in such case it is a good practice to simplify it. For example you could call this function in line 20.

pawel-peksa (changes requested):
Looks much better. Added one small nitpick

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,94 @@
...
+
+  const [title, setTitle] = useState<string>('');
+  const [description, setDescription] = useState<string>('');
+  const [imgUrl, setImgUrl] = useState<string>('');
+  const [imdbUrl, setImdbUrl] = useState<string>('');
+  const [imdbId, setImdbId] = useState<string>('');
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (isDisabled()) {
```
asakevych, line 20: linter don't like using isDisabled before initialization, try to move `handleSubmit`  under `isDisaabled` 😉

asakevych (changes requested):
Great job!!! 🚀 small change is needed for linter to pass 😉

choeqq (approved):
LGTM ✅

PR #1862, opened 2023-09-20
react_movies-list-add-form_v-01

DenysKolbasin:
[DEMO LINK](https://DenysKolbasin.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,94 @@
...
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const baseFormData = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
```
maxim2310, line 16: move it outside component

maxim2310 (changes requested):
Almost done! But need to make some improvements 

when i entered spaces in every input button still disabled but i dont see any error, need to add it, also need to add validation on url

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,125 @@
...
+    || !imdbUrl.trim()
+    || !imdbId.trim();
+
+  const handleMovieFieldChange = (name: string, value: string) => {
+    setNewMovie((prevMovie) => ({ ...prevMovie, [name]: value }));
+  };
+
+  function isValidUrl(url: string) {
+    if (!url) {
+      return undefined;
```
etojeDenys, line 37:
```suggestion
      return;
```
empty return by default returns undefined

etojeDenys (approved):
lgtm

PR #1874, opened 2023-09-28
Solution

Reject-4444:
https://reject-4444.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,50 +1,91 @@
-import { useState } from 'react';
+import { FormEvent, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = { onAdd: (movie: Movie) => void };
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,50 +1,91 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
VitaliyBondarenko1982, line 15: make one state as object for all input fields and use event.target.name for change proper field. attribute name in input should be the same to inputs state appropriate key
```
```
```suggestion
  const [newMovie, setNewMovie] = useState({
    title: '',
    ...
  });
```

VitaliyBondarenko1982 (changes requested):
Now I can input for each form field only spaces and can't to see any validate message, but add button still disabled. as user I don't know what happened.
Screenshot from 2023-09-28 12-14-59

plodnikate (approved):
Good job!

PR #1880, opened 2023-09-29
Add solution

JaneShavrukova:
[DEMO LINK](https://JaneShavrukova.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,91 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,91 @@
...
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```

```
const initialMovieState = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: ''
}
```

Example: 

```
}
```


```suggestion
  const [movie, setMovie] = useState(initialMovieState);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,91 @@
...
+
+  const resetForm = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
```
Viktor-Kost, line 23:
```suggestion
  const resetForm = () => {
    setMovie(initialMovieState);
  };
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,91 @@
...
+      imdbUrl,
+      imdbId,
+    });
+
+    setCount(prev => prev + 1);
+
+    resetForm();
+  };
+
+  const isSubmitDisabled = !(title
```
Viktor-Kost, line 41: I shouldn't be allowed to write only `spaces`
Screenshot 2023-09-30 at 09 50 52

Viktor-Kost (changes requested):

PR #1882, opened 2023-10-02
movies list

stunnerboyone:
[DEMO LINK](https://stunnerboyone.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,93 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (newMovie: Movie) => void;
+};
+
+export const NewMovie = ({ onAdd }: Props) => {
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,93 @@
...
+  onAdd: (newMovie: Movie) => void;
+};
+
+export const NewMovie = ({ onAdd }: Props) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):
let's show the error when I try to fill inputs with spaces

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,93 @@
...
+
+    setCount(count + 1);
+
+    setNewMovie({
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
```
anastasiia-tilikina, line 43: let's create a var for example `initialState` and reuse it to avoid code duplication

anastasiia-tilikina (changes requested):
just one more small fix

anastasiia-tilikina (approved):
Well done!

PR #1888, opened 2023-10-10
Solution

kostiantin-hudzenko:
[DEMO LINK](https://kostiantin-hudzenko.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,98 @@
-import { useState } from 'react';
+import classNames from 'classnames';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isButtonDisabled = () => {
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,53 +1,107 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
loralevitska, line 16: Combine it into one state
```suggestion
const [formValues, setFormValues] = useState({
  title: '',
  description: '', 
  imgUrl: '',
  imdbUrl: '',
  imdbId: '',
});
```

and then you could have only one handler 
```
const handleChange = ((value: string, name: string) => setFormValues(prevState => ({
  ...prevState,
  [name]: value,
})));
```

loralevitska (changes requested):
I can submit form with empty fields, fix it with [string.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim)

VitaliyBondarenko1982 (changes requested):
Good job.
Need to fix one problem.
Now I can fill all inputs with spaces and can't to see any error but submit button stiill disabled

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,53 +1,114 @@
...
+      imdbUrl,
+      imdbId,
+    } = formValues;
+
+    return !(title.trim() && imgUrl.trim()
+      && imdbUrl.trim() && imdbId.trim());
+  };
+
+  const handleChange = (value: string, name: string) => {
+    setFormValues((prevState) => ({
```
loralevitska, line 33: You could take field name from event target 

```suggestion
  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
  const { value, name } = e.target
    setFormValues((prevState) => ({
```

loralevitska (approved):

PR #1893, opened 2023-10-12
React_Movies-List-Add-Form

levchenko-dmytro:
[DEMO LINK](https://levchenko-dmytro.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={moviesList} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie
+          onAdd={(movie: Movie) => {
+            setMoviesList([...moviesList, movie]);
+          }}
```
loralevitska, line 20: Move to separate function  (transfer the function to a variable, and here only write the name of the function)

PR #1907, opened 2023-10-19
solution

zvir91:


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,109 @@
...
   // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  // to reset touched status of all the `Field`
+  const [count, setCount] = useState(0);
+  const [newTitle, setNewTitle] = useState('');
+  const [newDescription, setNewDescription] = useState('');
+  const [newImageURL, setNewImageURL] = useState('');
+  const [newURL, setNewURL] = useState('');
+  const [newId, setNewId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):
i can create the film only with spaces, let's fix it using the `trim`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,143 @@
...
+  const [count, setCount] = useState(0);
+  const [newMovie, setNewMovie] = useState(
+    {
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+
+    },
```
VitaliyBondarenko1982, line 21: write in constant object of initial state for new movie outside of component and use in both places (reset handler also)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,143 @@
...
+  };
+
+  const {
+    title,
+    imgUrl,
+    imdbUrl,
+    imdbId,
+  } = newMovie;
+
+  const disable = title && imgUrl && imdbUrl && imdbId;
```
VitaliyBondarenko1982, line 41: now i can fill all inputs with spaces, can't to see any error, add button not disabled but cant to add film.
if at least one required field not filled button should be disabled.
and trim value in TextField component when you count hasError.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +146,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!disable}
```
VitaliyBondarenko1982, line 149: it is hard to read (disabled settled not disable)
variable disable (better naming `isDisabled`) should be true when button need to disable.

VitaliyBondarenko1982 (changes requested):
Please check comments.

VitaliyBondarenko1982 (approved):
Looks good

PR #1914, opened 2023-10-20
Develop

YaroslavYarynych:
[DEMO LINK](https://YaroslavYarynych.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,20 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [add, onAdd] = useState<Movie[]>(moviesFromServer);
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={add} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie) => onAdd(([...add, movie]))} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,85 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (value: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const [state, setState] = useState({
+    title: ''.trim(),
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+
+  function handleReset(event: React.FormEvent) {
+    event.preventDefault();
+
+    setState({
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
+
+    onAdd(state);
+    setCount((prev) => prev + 1);
+  }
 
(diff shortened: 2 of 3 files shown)


Drach-Illya (approved):
Looks good to me.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,98 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+
+  const [state, setState] = useState({
+    title: ''.trim(),
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
```
polosanya, line 18: Same here. Also, it is better to create a separate state for each property

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,98 @@
...
+  const [errorImdbUrl, setErrorImdbUrl] = useState('');
+
+  function handleReset(event: React.FormEvent) {
+    event.preventDefault();
+
+    const newErrorImgUrl = !pattern.test(state.imgUrl) ? 'error' : '';
+    const newErrorImdbUrl = !pattern.test(state.imdbUrl) ? 'error' : '';
+
+    setErrorImgUrl(newErrorImgUrl);
+    setErrorImdbUrl(newErrorImdbUrl);
```
polosanya, line 33: https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

src/components/TextField/TextField.tsx
```diff
@@ -55,6 +58,10 @@ export const TextField: React.FC<Props> = ({
       {hasError && (
         <p className="help is-danger">{`${label} is required`}</p>
       )}
+
+      {(errorImgUrl || errorImdbUrl) && (
```
polosanya, line 62: Would be great to create a variable for this condition

src/App.tsx
```diff
@@ -1,16 +1,20 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [add, onAdd] = useState<Movie[]>(moviesFromServer);
```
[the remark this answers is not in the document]
YaroslavYarynych: If we change 'add' to 'movies'. tests won`t pass.

StasSokolov (changes requested):
Awesome job, let's improve your code a biit more🔥

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,122 @@
...
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+
+  const [hasImgUrl, setHasImgUrl] = useState('');
+  const [hasImdbUrl, setHasImdbUrl] = useState('');
+
+  const conditionForButton = !state.title
```
vasylmiskiv, line 25: Should write more appropriate name of the variable like `isButtonDisabled`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,122 @@
...
       <TextField
         name="imdbUrl"
         label="Imdb URL"
-        value=""
+        value={state.imdbUrl}
+        hasImdbUrl={hasImdbUrl}
+        onChange={value => {
+          setState({ ...state, imdbUrl: value });
+          setHasImdbUrl('');
+        }}
```
vasylmiskiv, line 110: It would be better if the error disappeared when the user entered the correct value in the input field instead of every new onChange event. Also, a method that resets the error on each text input is called every time

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,122 @@
...
+  }
+
+  function handleReset(event: React.FormEvent) {
+    event.preventDefault();
+
+    if (!pattern.test(state.imgUrl) && !pattern.test(state.imdbUrl)) {
+      setHasImgUrl('error');
+      setHasImdbUrl('error');
+      return;
+    }
```
vasylmiskiv, line 49: If you didn't write return in every if statement you wouldn't need to double check the validation of the inputs and then separately

vasylmiskiv (approved):

PR #1920, opened 2023-10-21
branch in development (uploading to check tests)

Drach-Illya:
[DEMO LINK](https://Drach-Illya.github.io/react_movies-list-add-form/)

Please be aware that this is an unfinished version. As the command 'npm test' always times out, I'm making this pull request to check tests and see what I need to improve. However, feel free to add feedback if you wish.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,36 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
 
+interface Movie{
+  title: string;
+  description: string;
+  imgUrl: string;
+  imdbUrl: string;
+  imdbId: string;
+}
+
 export const App = () => {
+  const [listOfMovies, setListOfMovies] = useState(moviesFromServer);
+
+  function addMovie(movie: Movie) {
+    const expandedList = [...listOfMovies, movie];
+
+    setListOfMovies(expandedList);
+  }
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={listOfMovies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie) => {
+          addMovie(movie);
+        }}
+        />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,107 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+interface Movie{
+  title: string;
+  description: string;
+  imgUrl: string;
+  imdbUrl: string;
+  imdbId: string;
+}
+
+type Add = (value: Movie) => void;
+
+export const NewMovie = ({ onAdd }: { onAdd: Add }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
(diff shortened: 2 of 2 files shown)


src/App.tsx
```diff
@@ -1,16 +1,36 @@
...
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={listOfMovies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie) => {
+          addMovie(movie);
+        }}
```
StasSokolov, line 32:
```suggestion
        <NewMovie onAdd={addMovie}
```

can be simplify

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,107 @@
...
+    setCount(count + 1);
+    setTitleValue('');
+    setDescriptionValue('');
+    setImgUrlValue('');
+    setImdbUrlValue('');
+    setImdbIdValue('');
+  };
+
+  const disabledCondition
+  = titleValue === ''
```
StasSokolov, line 42:
```suggestion
  = !titleValue
```
can be simplify, check please below also

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,107 @@
...
+  const movie: Movie = {
+    title: titleValue,
+    description: descriptionValue,
+    imgUrl: imgUrlValue,
+    imdbUrl: imdbUrlValue,
+    imdbId: imdbIdValue,
+  };
+
+  const reset = () => {
+    setCount(count + 1);
```
StasSokolov, line 33: is this correct? when user click reset button the count is increase? i think should be initial value

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,107 @@
...
+    || imdbUrlValue === '';
 
   return (
-    <form className="NewMovie" key={count}>
+    <form
+      className="NewMovie"
+      key={count}
+      onSubmit={() => {
+        onAdd(movie);
+        reset();
```
StasSokolov, line 53: in this place when user click add button the count should be increase

StasSokolov (changes requested):
Great job, let's improve your code a bit more!

src/components/NewMovie/NewMovie.tsx
Drach-Illya: // Increase the count after successful form submission

Yes, the comment in Master branch requests this to for reset

src/components/NewMovie/NewMovie.tsx
Drach-Illya: And it does in the 'Reset' function

Drach-Illya: In the comment Above

src/App.tsx
Drach-Illya: I don't think this is valid as we need to Recieve a `movie` value from a prop. 
Writing it your way will make it hard to recieve the Object with info about movie from the component NewMovie

image

src/components/NewMovie/NewMovie.tsx
Drach-Illya: Also, there is no reset button. Reset is done on the press of submit to clear the fields, which also ticks up count to prevent error message from appearing.

src/App.tsx
StasSokolov: ok, let's leave it!

Drach-Illya:
The `reset` function is tied to the `Submit` button, to reset the values in the fields to an empty string and increase the count to prevent error messages from showing up when we submit the movie. There is no dedicated `Reset` button on the website.

PR #1921, opened 2023-10-21
solution

vshepel:
[DEMO LINK](https://vshepel.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,25 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+
+import { Movie } from './types/Movie';
+
 import moviesFromServer from './api/movies.json';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+  const handleAddMovie = (movie: Movie) => {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,16 +1,25 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+
+import { Movie } from './types/Movie';
+
```
ivbutsykin, line 7: I suggest to remove this spaces because don't see huge sence in them:
```suggestion
import { Movie } from './types/Movie';
```

src/App.tsx
```diff
@@ -1,16 +1,25 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+  const handleAddMovie = (movie: Movie) => {
+    setMovies([...movies, movie]);
```
ivbutsykin, line 13: Better to use "previous" value of state in such cases:
```suggestion
    setMovies(prevMovies => [...prevMovies, movie]);
```

ivbutsykin (approved):
Good job!

PR #1931, opened 2023-10-22
Completed the task

Maksim114488:
https://Maksim114488.github.io/react_movies-list-add-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
+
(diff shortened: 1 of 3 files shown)


mpohorenyi (approved):
Good job!
If the link is not valid, it still allows you to add the card to the list, although this should not be the case

superpooperxxx (changes requested):
Fix comment of your mate🤗

superpooperxxx (approved):
Awesome🔥

PR #1934, opened 2023-10-23
add task solution

Vovchukkul:
https://Vovchukkul.github.io/react_movies-list-add-form/

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ You have the `App` with the `MoviesList` and `NewMovie` form containing ready
 to use `TextField` components. Learn how it works and implement an ability to
 add movies from [IMDB](https://www.imdb.com/).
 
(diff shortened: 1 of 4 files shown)


sTorba24 (changes requested):
Almost good! 

imdb id should be required
image

also i should be able to add `spaces`, just should be length using .trim(), spaces can be

sTorba24 (changes requested):
As i mentioned before only spaces shouldn't be allowed, use .trim() to check input values
image

Still able to add movie without imdb url

sTorba24 (changes requested):
Let user put spaces everywhere, just check inputValue.trim() in result
if no length - disable button, show errors. imdb id still not required
Fix it more properly
`    text` is ok, use trim in result
`     ` is not ok - show message that field required, block btn
image

etojeDenys (approved):
good job

PR #1935, opened 2023-10-23
task solution

valeriaoleynik:
[DEMO LINK](https://valeriaoleynik.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,113 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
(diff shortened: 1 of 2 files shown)


anastasiia-tilikina (changes requested):
well done! 

check my comments.
Also, now I'm able to add movies with spaces only in text fields. disable this behaviour

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,93 @@
...
+export const NewMovie: React.FC<Props> = ({
+  onAdd,
+}) => {
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

don't forget to specify the name attr to input in TextField

etojeDenys (changes requested):

nazarmatsevych (approved):
Awesome!

PR #1938, opened 2023-10-24
Develop

atrofymovych:

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,123 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   const [count] = useState(0);
 
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const [titleError, setTitleError] = useState(false);
+  const [imgUrlError, setImgUrlError] = useState(false);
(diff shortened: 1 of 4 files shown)


VitaliyBodnar (changes requested):
- demo link is not valid 
Screenshot 2023-10-24 at 13 14 52

atrofymovych:
fixed

nazarmatsevych (changes requested):
I can add movie with empty fields, use trim to remove extra spaces
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,130 @@
...
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   const [count] = useState(0);
 
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,123 @@
...
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   const [count] = useState(0);
```
denys-danyliuk, line 10:
```suggestion
  const [count, setCount] = useState(0);
```

denys-danyliuk: you need to change it's value after you'll add a new movie

denys-danyliuk (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,124 @@
...
+        imgUrlError: true,
+        imdbUrlError: true,
+        imdbIdError: true,
+      });
+
+      return;
+    }
+
+    onAdd(newMovie);
+    setCount(count + 1);
```
denys-danyliuk, line 59:
```suggestion
    setCount(prev => prev + 1);
```
if you compute new value depending on old one you should use callback

denys-danyliuk (approved):

PR #1939, opened 2023-10-24
✼

Masha-Cactus:


xiimiik (changes requested):
i can add bot a url to url field, it shouldnt be like that
image

xiimiik (approved):
my bad

PR #1940, opened 2023-10-24
task solution

DenysShurubura:
[DEMO LINK](https://DenysShurubura.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,95 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+import { Movie } from '../../types/Movie';
+
+interface Props {
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
+
+  const submitHandler = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    resetInputs();
+    setCount(count + 1);
```
superpooperxxx, line 29: Not a big problem here, but I suggest moving this calls after `onAdd`, cause you should add first and then reset

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+      imdbId,
+    });
+  };
+
+  const validator = () => {
+    if (title && imgUrl && imdbUrl && imdbId) {
+      return true;
+    }
+
+    return false;
```
superpooperxxx, line 45: You can just `return title && imgUrl && imdbUrl && imdbId`. Actually you don't even need this function, just save the store in variable

superpooperxxx (changes requested):
Good job🔥 Little improvements

superpooperxxx (approved):
🔥

PR #1942, opened 2023-10-25
react movies list add form

whiovt:
[DEMO LINK](https://whiovt.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,27 +1,67 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,27 +1,67 @@
...
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
anastasiia-tilikina, line 15: let's combine these states into one object:

```typescript
const [newMovie, setNewMovie] = useState({ title: '', description: '', // ... })
``` 

also, let's create variable with initial state to reuse it on reset.

anastasiia-tilikina (changes requested):
I cannot add any movie since i cannot enter any text in imgUrl field.
Make sure that you've passed tests (at least most of them)

PR #1944, opened 2023-10-25
Develop

andrijdudar:
[DEMO LINK](https://andrijdudar.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,10 +1,67 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,67 @@
...
+
+  const [title, setTitle] = useState('');
+
+  const [description, setDescription] = useState('');
+
+  const [imgUrl, setImgUrl] = useState('');
+
+  const [imdbUrl, setImdbUrl] = useState('');
+
+  const [imdbId, setImdbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):

IvanFesenko (approved):
Well done

PR #1949, opened 2023-10-30
react_movies-list-add-form solution 

Medaed:
[DEMO LINK](https://Medaed.github.io/react_movies-list-add-form/)


nazarmatsevych (changes requested):
I can add movie with empty fields, use trim() method to fix it:
image

SanyaBratashchuk (approved):

PR #1950, opened 2023-10-30
add task solution

mykyta01:


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,97 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):
i can create the movie only with spaces let's fix that

etojeDenys (changes requested):
still not fixed
image

src/App.tsx
```diff
@@ -1,16 +1,26 @@
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+
+const initialMovies: Movie[] = moviesFromServer.map(movies => ({ ...movies }));
```
Andriy-Shymkiv, line 8: what's the purpose of this mapping? you take movies and for each movie return the same movie. it doesn't make any sense, isn't it?

src/components/NewMovie/NewMovie.tsx
Andriy-Shymkiv: not fixed at all ⬆️

Andriy-Shymkiv (changes requested):

anastasiiavorobiova (changes requested):
Great progress here! Consider fixing the last issue
Screenshot 2023-11-04 at 10 29 49

Screenshot 2023-11-04 at 10 29 34

mykyta01:
> Great progress here! Consider fixing the last issue Screenshot 2023-11-04 at 10 29 49
> 
> Screenshot 2023-11-04 at 10 29 34

mykyta01:
Screenshot 2023-11-04 at 09 41 58
For my view everything works fine

sTorba24 (changes requested):
It's not good practice to let submit form with just `spaces` if you have restriction for empty string, it's almost the same

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,90 @@
...
+  const [newMovie, setNewMovie] = useState({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+
+  const invalid = newMovie.title
+  && newMovie.imgUrl && newMovie.imdbUrl && newMovie.imdbId;
```
sTorba24, line 21:
```suggestion
  const invalid = newMovie.title.trim()
  && newMovie.imgUrl.trim() && newMovie.imdbUrl.trim() && newMovie.imdbId.trim();
```

sTorba24 (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,92 @@
...
+
+  const [newMovie, setNewMovie] = useState({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+
+  const invalid = newMovie.title.trim()
```
sTorba24, line 20:
```suggestion
  const valid = newMovie.title.trim()
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +95,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!invalid}
```
sTorba24, line 98:
```suggestion
            disabled={!valid}
```

src/components/NewMovie/NewMovie.tsx
sTorba24: because true value means here that all fields has value(so form valid)

sTorba24 (approved):

PR #1958, opened 2023-11-01
implementaion

Daniilart01:
[DEMO LINK](https://Daniilart01.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,128 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd(movie: Movie): void,
+}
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,128 @@
...
+
+  const [isButtonDisabled, setIsButtonDisabled] = useState(true);
+  const [isImdbUrlWrong, setIsImdbUrlWrong] = useState(false);
+  const [isImgUrlWrong, setIsImgUrlWrong] = useState(false);
+
+  const [description, setDescription] = useState('');
+  const [title, setTitle] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
etojeDenys, line 20: we can combine it in one useState
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):

Daniilart01:
Good idea, I was thinking about groupping this states into one, but I didn`t think about probability of setting name attr and passing to handleChange `event` insteaad of `string`

lerastarynets (approved):
You're doing amazing, keep it up!

PR #1961, opened 2023-11-05
solution

KseniyaGonchar:
https://KseniyaGonchar.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,105 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,105 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
VitaliyBondarenko1982, line 17:
```suggestion
  const [movi, setMovie] = useState({
     title: '',
     description: '',
    ...
  });
```
make one state for all form fields and one handler for change it. use event.target.name for change proper field of movie.
for it name attribute in input should have same value as movie field.

VitaliyBondarenko1982 (changes requested):
Good job.
Needs some improvements.

Now i can fill some required fields only wiht spaces and can't see any error but add button still disabled. show validation error for trimmed input value.
 
Screenshot from 2023-11-05 17-46-16

VitaliyBondarenko1982 (approved):
Well done

PR #1969, opened 2023-11-08
add task solution

markiuan:
[DEMO LINK](https://markiuan.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,121 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface NewMovieProps {
+  onAdd: (movie: Movie) => void;
+}
+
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,121 @@
...
+      case 'imdbUrl':
+        setImdbUrl(value);
+        break;
+      case 'imdbId':
+        setImdbId(value);
+        break;
+      default:
+        break;
+    }
+  };
```
nazarmatsevych, line 38: we can combine it in one useState

```
  title: '',
  ...
})
```

```
}
```
don't forget to specify name attr to input in TextField

nazarmatsevych (changes requested):
I can add movie with empty fields, use trim() method to fix it
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,94 @@
...
+  const [newMovie, setNewMovie] = useState<Movie>({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+
+  const handleChange = (name: string, value: string) => {
+    // const { name, value } = e.target;
```
etojeDenys, line 21:
```suggestion
```

etojeDenys (approved):
good job

PR #1973, opened 2023-11-13
Solution

Vasyl-Zhyliakov:
[Demo](https://Vasyl-Zhyliakov.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,27 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = (movie: Movie) => {
+    setMovies((prev) => [...prev, movie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,98 @@
...
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
+  };
+
+  const handleInputChange
+    = (event: React.ChangeEvent<HTMLInputElement>, field: string) => {
+      setNewMovie({ ...newMovie, [field]: event.target.value });
```
TarasHoliuk, line 40:
```suggestion
      setNewMovie({ ...newMovie, [event.target.name]: event.target.value });
```

you could do it in this way - get input name from the event, instead of `field` argument - FYI

src/App.tsx
```diff
@@ -1,16 +1,27 @@
...
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie: Movie) => {
+          addMovie(movie);
+        }}
+        />
```
TarasHoliuk, line 24:
```suggestion
        <NewMovie onAdd={addMovie} />
```
`addMovie` is a function that accepts `movie` argument. No need to add additional function wrapper, just use reference to `addMovie` function

FYI only

TarasHoliuk (approved):
Very good job, I like it :+1: 

Added 2 comments just FYI (take into account in the future)

PR #1984, opened 2023-11-20
Solution

IrynaK-a:
[DEMO LINK](https://IrynaK-a.github.io/react_movies-list-add-form/)


anastasiiavorobiova (changes requested):
Great start! Consider trimming values, a user shouldn't be able to add film with spaces only

witflash (approved):
Well done 👍🏻

PR #1993, opened 2023-11-24
made solution for list add form

MrHunterBoi:
[DEMO LINK](https://MrHunterBoi.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,90 @@
 import { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd: addMovie }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgURL] = useState('');
+  const [imdbUrl, setImdbURL] = useState('');
+  const [imdbId, setImdbID] = useState('');
+
+  const reset = () => {
+    setTitle('');
+    setDescription('');
+    setImgURL('');
+    setImdbURL('');
+    setImdbID('');
+  };
+
+  const handleSubmit = () => {
+    setCount(count + 1);
+    addMovie({
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    });
+
+    reset();
+  };
+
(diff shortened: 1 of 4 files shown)


yevtok (approved):
Nice job!

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,92 @@
...
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd: addMovie }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgURL] = useState('');
+  const [imdbUrl, setImdbURL] = useState('');
+  const [imdbId, setImdbID] = useState('');
```
BudnikOleksii, line 15: You can create one state for all fields, it should be an object with these properties, also you can create a default object that can be used for reset this state:
```javascript
{
  title: '',
  // other
}
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,92 @@
...
+      imdbUrl,
+      imdbId,
+    });
+
+    reset();
+  };
+
+  const checkIfUrlValid = (url: string) => {
+    // eslint-disable-next-line max-len
+    const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
```
BudnikOleksii, line 42: Move this pattern outside, you can create a constant for regex, also a name should describe what this pattern actually validate

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +95,12 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={
+              title.trim() === ''
+              || imgUrl.trim() === ''
+              || imdbUrl.trim() === ''
+              || imdbId.trim() === ''
```
BudnikOleksii, line 102: Create a variable for this condition

BudnikOleksii (changes requested):
Good work 👍 
left a few suggestions
Also, the app works a bit strange. I entered url/image, but recieve a notification that these fields required 🤷‍♂️  And with these errors I still can submit the form:
image

MrHunterBoi:
Ah, right, forgot to add a condition for posting xd

BudnikOleksii (approved):
Nicely done 👍 
The message about validation is incorrect, "Image URL is required" even user pass something. In this case, it should be invalid instead of required.

PR #1994, opened 2023-11-24
solution

yevtok:
[DEMO LINK](https://yevtok.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +90,7 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={isSubmitActive}
```
ArtemTeslenko, line 93: disabled when active? you can call this function `isSubmitDisabled` for example

ArtemTeslenko (approved):
Great 🔥 just one comment 👍

PR #1998, opened 2023-11-28
first commit

Daniil-Lilin:
[demo link](https://Daniil-Lilin.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,8 +9,9 @@ import { Movie } from './types/Movie';
 export const App = () => {
   const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
 
-  const addMovie
-    = (movie: Movie) => setMovies(prevMovies => [...prevMovies, movie]);
+  const addMovie = (movie: Movie) => setMovies(
+    prevMovies => [...prevMovies, movie],
+  );
 
   return (
     <div className="page">
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -55,6 +55,19 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
     setCount(prevCount => prevCount + 1);
   };
 
+  const checkIfDisabled = () => {
+    if (
+      !title.trim()
+      || !imdbId.trim()
+      || !imdbUrl.trim()
+      || !imgUrl.trim()
+    ) {
+      return true;
+    }
+
+    return false;
+  };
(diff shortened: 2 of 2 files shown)


etojeDenys (changes requested):
i can create the film only with spaces, let's fix that
image

src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  const addMovie
+    = (movie: Movie) => setMovies(prevMovies => [...prevMovies, movie]);
```
volodymyr-soltys97, line 13:
```suggestion
  const addMovie = (movie: Movie) => setMovies(prevMovies => [...prevMovies, movie]);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +112,12 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={
+              !title.trim()
+              || !imdbId.trim()
+              || !imdbUrl.trim()
+              || !imgUrl.trim()
```
volodymyr-soltys97, line 119: Will be better if you create a variable for this checking and use it here

volodymyr-soltys97 (changes requested):
Almost done!
Let's improve your code

src/App.tsx
Daniil-Lilin: I can't make like this because of this string has more than 80 symbols

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,122 @@
...
+
       />
 
       <TextField
         name="imgUrl"
         label="Image URL"
-        value=""
+        value={imgUrl}
+        onChange={handleInputChange}
+        regEx={validImgUrl}
```
l4st1m0za, line 100:
```suggestion
        isRegExValid={validImgUrl}
```
It's better to change the name accordingly so that it means what it has inside. As the name **regEx** means that it will contain some kind of a regEx, but you pass **boolean** insted

l4st1m0za (approved):
Great job! But please, don't forget to make correct naming!

PR #2006, opened 2023-12-01
add task solution

Softjey:
I also implemented URL validation and to make it look better I reworked the textField a bit, so the tests don't pass, but everything looks even better
[DEMO LINK](https://softjey.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,98 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+interface Props {
+  onAdd: (movie: Movie) => void,
+
+}
+
+export const NewMovie: React.FC<Props> = ({
+  onAdd,
+}) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const { 0: count, 1: setCount } = useState(0);
+
+  const { 0: title, 1: setTitle } = useState('');
+  const { 0: description, 1: setDescription } = useState('');
+  const { 0: imgUrl, 1: setImgUrl } = useState('');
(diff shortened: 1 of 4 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
There are now options to add empty values as a movie, this needs to be fixed
image
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,109 @@
...
+export const NewMovie: React.FC<Props> = ({
+  onAdd,
+}) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const { 0: count, 1: setCount } = useState(0);
+  const { 0: hasSomeError, 1: setHasSomeError } = useState(false);
+
+  const { 0: title, 1: setTitle } = useState('');
```
etojeDenys, line 18:
```suggestion
  const [title, setTitle] = useState('');
```
we usually write use states like this

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,109 @@
...
+  const { 0: description, 1: setDescription } = useState('');
+  const { 0: imgUrl, 1: setImgUrl } = useState('');
+  const { 0: imdbUrl, 1: setImdbUrl } = useState('');
+  const { 0: imdbId, 1: setImdbId } = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):

src/components/NewMovie/NewMovie.tsx
Softjey: Array Assignment Pattern is 30% slower than Object Assignment Pattern. This happens due to the peculiarity of converting JS code into byte code when using the Array Assignment Pattern
https://tc39.es/ecma262/#prod-ArrayAssignmentPattern
https://tc39.es/ecma262/#prod-ObjectAssignmentPattern

nazarmatsevych (approved):
Good work, please see my comment about useState syntax

PR #2022, opened 2023-12-05
add task solution

yuliiababych:
[DEMO LINK](https://yuliiababych.github.io/react_movies-list-add-form/)


lerastarynets (changes requested):
Pls add error text to inputs, because now user can't understand what exactly is wrong with the link
Screenshot at Dec 05 12-00-32

lerastarynets (approved):
I meant that you should add something like the following for user to understand what is wrong, so consider adding it
Screenshot at Dec 05 13-32-07

PR #2023, opened 2023-12-05
done task Movies List Add Form

Anton-Kuchmasov:
[DEMO LINK]
(https://Anton-Kuchmasov.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [moviesList, setMoviesList] = useState(moviesFromServer);
+
+  const addNewMovieToList = (movie: Movie) => {
+    setMoviesList([...moviesList, movie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={moviesList} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie) => addNewMovieToList(movie)} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,101 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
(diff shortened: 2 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,101 @@
...
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
+  const [movieTitle, setMovieTitle] = useState(emptyMovie.title);
+  const [movieDescription, setMovieDescription]
+    = useState(emptyMovie.description);
+  const [imgURL, setImgURL] = useState(emptyMovie.imgUrl);
+  const [imdbURL, setImdbURL] = useState(emptyMovie.imdbUrl);
+  const [imdbId, setImdbId] = useState(emptyMovie.imdbId);
```
ksyvvkaa, line 25: you could create just one state with an object where are all those fields and with `emptyMovie` object by default.
like this:
```
 const [newMovie, setNewMovie] = useState<Movie>(emptyMovie);
```
and then handle all inputs with setting new object to state with updated field: 
```
  const handleInputChange = (key: string) => {
    return (value: string) => setInputs({
      ...newMovie,
      [key]: value,
    });
  };
```

and then `onChange={handleInputChange('title')}`  for example

ksyvvkaa (changes requested):
Good work 👍 but fix this:

image

image

so if I enter nothing but spaces, I should'n be able to press submit button

src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={moviesList} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie) => addNewMovieToList(movie)} />
```
ksyvvkaa, line 21: I can't understand the purpose of passing this func `(movie) => addNewMovieToList(movie)` as `onAdd` here and not just `onAdd={addNewMovieToList}`

src/components/NewMovie/NewMovie.tsx
Anton-Kuchmasov: i'm totally agree with you,
it's can be shorter and (probably?) easier,

btw, in case of learning states and practicing i've chosen this, long and harder way)

src/App.tsx
Anton-Kuchmasov: fixed, thx)

Anton-Kuchmasov:
> Good work 👍 but fix this:
> 
> image
> 
> image
> 
> so if I enter nothing but spaces, I should'n be able to press submit button

added trim() to all fields

ksyvvkaa (approved):

polosanya (approved):
Well done! 👍

PR #2038, opened 2023-12-12
Develop

Patrykmclaren14:
[solution](https://Patrykmclaren14.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,10 +1,64 @@
-import { useState } from 'react';
+import { MouseEvent, SetStateAction, useState } from 'react';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
+interface NewMovieProps {
+  onAdd: (newMovie: Movie) => void;
+}
+
+interface Movie {
+  title: string,
+  description: string,
+  imgUrl: string,
+  imdbUrl: string,
+  imdbId: string,
+}
+
+export const NewMovie: React.FC<NewMovieProps> = ({ onAdd }) => {
   const [count] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const addMovie = (event: MouseEvent<HTMLButtonElement, globalThis.MouseEvent>) => {
+    event.preventDefault();
+
+    const newMovie = {
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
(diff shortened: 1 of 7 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -13,33 +59,40 @@ export const NewMovie = () => {
...
+        onChange={handleTitleAdd}
         required
       />
 
       <TextField
         name="description"
         label="Description"
-        value=""
+        value={description}
+        onChange={handleDescriptionAdd}
```
Barosz30, line 71: If you change that to 
value={description}
onChange={setDescription}
it should work and you can delete bunch of code above. For title and other states that also should work ;)

Barosz30 (approved):
Very good code. I apprecieate Your commits. Gives preety clear view on what's going on

src/components/NewMovie/NewMovie.tsx
mvjl000: Yeah, I guess it could work ;D

src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +101,13 @@ export const NewMovie = () => {
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            onClick={event => addMovie(event)}
+            disabled={
+              !title
+              || !imgUrl
+              || !imdbUrl
+              || !imdbId
+            }
```
mvjl000, line 110: I would create a constant for this

mvjl000 (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,37 @@
...
+      imdbId,
+    };
+
+    onAdd(newMovie);
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+    setCount(count + 1);
```
JoyCoffeeAddict, line 33: alwyas use prevState syntax to set state based on a previous value.

JoyCoffeeAddict (changes requested):

src/components/NewMovie/NewMovie.tsx
mbruhler: I won't agree with that, it's always good to create handler for event, not just update the state

mbruhler (changes requested):
Hi,

Overall good job, it works.
One thing to change is how you handle form.

Please create 
```
//somewhere in deep galaxy above component
const EMPTY_FORM = {
    title: ''
    description: ''
    imgUrl: ''
    imdbUrl: ''
    imdbId: ''
}
//in component, we create new state form to store whole form there
const [form, setForm] = useState<Form>(EMPTY_FORM);

//works itself for onAdd as onAdd(form)
//when emptying the form, use setForm(EMPTY_FORM)
```

JoyCoffeeAddict (approved):

PR #2045, opened 2023-12-14
solution

bertsu02:
[DEMO LINK](https://bertsu02.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,95 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (movie: Movie) => void
+}
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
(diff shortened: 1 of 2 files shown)


zaura333 (approved):
Looks good!

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
+}
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const invalid = !title || !imgUrl || !imdbUrl || !imdbUrl || !imdbId;
```
mbruhler, line 16: `imdbUrl` is twice

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,95 @@
...
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (movie: Movie) => void
+}
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
```
mbruhler, line 12: It's better to use some sort of FORM object rather than keeping state for every field separately

mbruhler (changes requested):
Hi,
in overall good job, BUT

- you keep state for each field separately and what's worse, you assign these separate values back to JSON object when sendind data up to onAdd. These values should be stored in JSON state object, like form or something. 
- in you invalid field in NewMovie you write imdbUrl twice

Fix these issues and we're good to go

mbruhler (approved):
Great, one minor tip is that you could store 
```
{
      title: '',
      description: '',
      imgUrl: '',
      imdbUrl: '',
      imdbId: '',
    }
```
above the component and use it when creating state and when emptying the form

PR #2047, opened 2023-12-15
Develop

szymanski-marcin:
[DEMO LINK](https://szymanski-marcin.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,99 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [count, setCount] = useState(0);
+
+  const data = {
+    title: title.trim(),
+    description: description.trim(),
+    imgUrl: imgUrl.trim(),
+    imdbUrl: imdbUrl.trim(),
+    imdbId: imdbId.trim(),
+  };
+
+  const isDataValid = Boolean(
+    data.title
+    && data.imgUrl
+    && data.imdbUrl
+    && data.imdbId,
+  );
+
+  const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    if (!isDataValid) {
+      return;
+    }
+
+    onAdd(data);
+
+    setCount(prevCount => prevCount + 1);
+
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
 
   return (
-    <form className="NewMovie" key={count}>
+    <form
+      className="NewMovie"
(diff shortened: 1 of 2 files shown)


mosowiecka (approved):
Looks great.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,92 @@
...
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [title, setTitle] = useState('');
```
mbruhler, line 10: it's better to use JSON to store every property of the form together

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,92 @@
...
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [count, setCount] = useState(0);
+
+  const data = {
```
mbruhler, line 17: And if you will use Form as a state object, then you wouldn't need this, because everything up-to-date would be in this state form object.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,92 @@
...
+      onSubmit={handleSubmit}
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
```
mbruhler, line 56: using form state object, here it would be form.title for example

mbruhler (changes requested):
Hi,

Overall good job, it works.
One thing to change is how you handle form.

Please create 
```
//somewhere in deep galaxy above component
const EMPTY_FORM = {
    title: ''
    description: ''
    imgUrl: ''
    imdbUrl: ''
    imdbId: ''
}
//in component, we create new state form to store whole form there
const [form, setForm] = useState<Form>(EMPTY_FORM);

//works itself for onAdd as onAdd(form)
//when emptying the form, use setForm(EMPTY_FORM)
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,89 @@
...
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={form.title}
+        onChange={(value) => setForm({ ...form, title: value })}
```
JoyCoffeeAddict, line 54: make a handler for it

JoyCoffeeAddict (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,96 @@
...
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={form.title}
+        onChange={handleTitleChange}
```
mbruhler, line 61: Not only for this one! For every onChange event, it applies to every FORM field

mbruhler (changes requested):
Apply handlers for all form fields

src/components/NewMovie/NewMovie.tsx
mbruhler: You can pass the first argument as field name and second value for this field

mbruhler: `handleChange('title', e.target.value)`
`handleChange('imdbUrl', e.target.value)`

or you can use currying to do it faster

`const handleChange = (key: keyof Movie) => (
    eventText: string,
  ) => setNewMovie({
    ...newMovie,
    [key]: eventText,
  });`

mbruhler (approved):

PR #2056, opened 2023-12-20
Solution

web3-sorcerer:
[DEMO LINK](https://web3-sorcerer.github.io/react_movies-list-add-form/)


src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState([...moviesFromServer]);
+
+  const onAdd = (movie: Movie) => {
+    setMovies(prev => [...prev, movie]);
+  };
```
mvjl000, line 13: You need some additional validation for adding movie.
Disabling the button isn't always enough. User can remove the `disabled` attribute from the button (via devtools) and this can cause your app to crash

mvjl000 (changes requested):
Very good overall, but there is a place for significant security improvement ;)

mpiskadlo (changes requested):
Sorry, wrong account previously 😆

web3-sorcerer:
I hope that my fix is enough 🤞

aureliagorzna (approved):

mbruhler (approved):

PR #2059, opened 2023-12-21
created form for add new movie

ivan-baranovskyi:
[DEMO LINK](https://ivan-baranovskyi.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,114 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setimgUrl] = useState('');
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,114 @@
...
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setimgUrl] = useState('');
+  const [imdbUrl, setimdbUrl] = useState('');
+  const [imdbId, setimdbId] = useState('');
+
+  // const [disabledButton, setDisabledButton] = useState(true);
```
volodymyr-soltys97, line 20: Remove comments everywhere
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Need to fix it
image
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,111 @@
...
+  onAdd: (movie: Movie) => void,
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
```
Viktor-Kost, line 14: It's not advisable to generate such a large number of states when we only have one instance (movie). 

```
const initialMovieState = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: ''
}
```

Example: 

```
}
```

```suggestion
  const [movie, setMovie] = useState(initialMovieState);
```

Viktor-Kost (changes requested):

volodymyr-soltys97 (approved):
Great work 🔥

PR #2062, opened 2023-12-22
add form

Vlad4567:
- [DEMO LINK](https://Vlad4567.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -2,44 +2,81 @@ import { useState } from 'react';
 import { TextField } from '../TextField';
 
 export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isSubmitDisable = !(
+    title.trim()
+    && imgUrl.trim()
+    && imdbUrl.trim()
+    && imdbId.trim()
+  );
+
+  const handleSubmit = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -2,44 +2,81 @@ import { useState } from 'react';
...
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
maxim2310, line 11: better to use one state with object

src/components/NewMovie/NewMovie.tsx
```diff
@@ -2,44 +2,81 @@ import { useState } from 'react';
...
+    && imdbUrl.trim()
+    && imdbId.trim()
+  );
+
+  const handleSubmit = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
```
maxim2310, line 25: and need to create initial movie object for using here for clean form and as initial value in state

maxim2310 (changes requested):

src/components/TextField/TextField.tsx
```diff
@@ -23,13 +24,15 @@ export const TextField: React.FC<Props> = ({
   placeholder = `Enter ${label}`,
   required = false,
   onChange = () => {},
+  pattern,
 }) => {
   // generage a unique id once on component load
```
OleziO, line 29: Remove all comments from the code

OleziO (changes requested):
After filling out the form, the film is not added to the list of films

VitaliyBondarenko1982 (changes requested):
As user I filled form. See some fields with red color but without any validation error. Can add film. But after click nothing happened.
Please check your logic (show some error text if some field is invalid, keep add button disabled if some field is invalid).
chrome-capture-2023-12-23

volodymyr-soltys97 (approved):
Great work 🔥

PR #2064, opened 2023-12-25
movies adding form

baraban2003:
[DEMO LINK](https://baraban2003.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,87 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,87 @@
...
+type Props = {
+  onAdd: (movie: Movie) => void
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setimdbId] = useState('');
```
etojeDenys, line 14: we can combine it in one useState
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):
i can create the movie only with spaces
image
let's fix that

src/components/NewMovie/NewMovie.tsx
baraban2003: Tnx.
It was not easy (
The previous decision I liked more, It looked easier.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,91 @@
...
+    setNewMovie(
+      defaultMovie,
+    );
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    onAdd(newMovie);
+    setCount(count + 1);
```
l4st1m0za, line 38:
```suggestion
    setCount(count++);
```

l4st1m0za (approved):
Great job!

src/components/NewMovie/NewMovie.tsx
baraban2003: Right. Tnx

PR #2069, opened 2024-01-01
add task solution

dimakrugly:
[DEMO LINK](https://dimakrugly.github.io/react_movies-list-add-form/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Need to use trim for all fields before sending
It will not be possible to add empty fields
image
image

lerastarynets (approved):
lgtm 🚀

PR #2072, opened 2024-01-04
Develop

boytrend0108:
[DEMO LINK](https://boytrend0108.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,113 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Inputs } from '../../types/Inputs';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (newMovie: Movie) => void
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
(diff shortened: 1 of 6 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,113 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):
1. i can create the movie only with spaces
image
let's fix that

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,98 @@
...
+        value={newMovie.title}
+        onChange={event => handleOnChange(event)}
       />
 
       <TextField
         name="description"
         label="Description"
-        value=""
+        value={newMovie.description}
+        onChange={event => handleOnChange(event)}
```
lerastarynets, line 73:
```suggestion
        onChange={handleOnChange}
```
**PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

lerastarynets (changes requested):
The errors are shown for empty fields however I'm still able to add a movie with all fields consisting of whitespaces pls fix

anastasiiavorobiova (approved):
Great work!

PR #2073, opened 2024-01-05
react with ts

nazardudnyk:
-- [DEMO LINK](https://nazardudnyk.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,26 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+
+const initialMovies: Movie[] = moviesFromServer.map(value => ({ ...value }));
 
 export const App = () => {
+  const [movies, setMovies] = useState(initialMovies);
+
+  const addHandler = (movie: Movie) => {
+    setMovies(prevMovies => [...prevMovies, movie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addHandler} />
       </div>
     </div>
   );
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,16 +1,26 @@
...
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+
+const initialMovies: Movie[] = moviesFromServer.map(value => ({ ...value }));
 
 export const App = () => {
+  const [movies, setMovies] = useState(initialMovies);
+
+  const addHandler = (movie: Movie) => {
```
Viktor-Kost, line 13: more correct name
```suggestion
  const addMovie = (movie: Movie) => {
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,85 @@
...
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```

```
const initialMovieState = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: ''
}
```

Example: 

```
}
```

```suggestion
  const [movie, setMovie] = useState(initialMovieState);
```

Viktor-Kost (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,94 @@
...
+
+  const clear = () => {
+    setNewMovie({
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
+  };
```
Viktor-Kost, line 36:
```suggestion

const initialMovieState = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: ''
}

export const NewMovie: React.FC<Props> = ({ onAdd }) => {
  const [count, setCount] = useState(0);

  const [newMovie, setNewMovie] = useState<Movie>(initialMovieState);

  const {
    title, description, imgUrl, imdbUrl, imdbId,
  } = newMovie;

  const disabled = !(
    title.trim() && imgUrl.trim() && imdbUrl.trim() && imdbId.trim()
  );

  const clear = () => {
    setNewMovie(initialMovieState);
  };
```

Viktor-Kost (approved):
Nice!

PR #2075, opened 2024-01-06
react_movies-list-add-form

serjgrm:
[DEMO LINK](https://serjgrm.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,29 +1,21 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { TextField } from '../TextField';
-
-export interface MoviesProps {
-  title: string;
-  description: string;
-  imgUrl: string;
-  imdbUrl: string;
-  imdbId: string;
-}
+import { Movie } from '../../types/Movie';
 
 interface NewMovieProps {
-  setMovies: React.Dispatch<React.SetStateAction<MoviesProps[]>>;
+  onAdd: React.Dispatch<React.SetStateAction<Movie[]>>;
 }
 
-export const NewMovie: React.FC<NewMovieProps> = ({ setMovies }) => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
+export const NewMovie: React.FC<NewMovieProps> = ({ onAdd }) => {
   const [title, setTitle] = useState('');
   const [description, setDescription] = useState('');
   const [imgUrl, setImgUrl] = useState('');
   const [imdbUrl, setImdbUrl] = useState('');
   const [imdbId, setImdbId] = useState('');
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,101 @@
...
+interface NewMovieProps {
+  onAdd: React.Dispatch<React.SetStateAction<Movie[]>>;
+}
+
+export const NewMovie: React.FC<NewMovieProps> = ({ onAdd }) => {
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
VitaliyBondarenko1982, line 14:
```suggestion
  const [newMovie, setNewMovie] = useState({
     title: '',
     description: '',
     ...
  });

```
suggest to make one state for all form fields and use event.target.name for change proper newMovie field.
required to set appropriate name attributes for inputs.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,101 @@
...
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
+
+  useEffect(() => {
+    if (title && imdbId && imdbUrl && imgUrl) {
+      setDisabled(false);
+    }
+  }, [title, imdbUrl, imgUrl, imdbId]);
```
VitaliyBondarenko1982, line 30: it is redundant to make state for such codition. enough to make variable. and looks like it is should be with || logical operator, as button should be disabled when some required input is empty.
trim should solve problem in screencast (now possible add movie with only spaces in input)

```suggestion
 const isDisabled = title.trim() ||  imdbId.trim() || imdbUrl.trim() || imgUrl.trim()
```
chrome-capture-2024-1-6
 
Screenshot from 2024-01-06 17-41-55

VitaliyBondarenko1982 (changes requested):
Needs some improvements.
Please check comments.

src/components/TextField/TextField.tsx
```diff
@@ -22,15 +24,26 @@ export const TextField: React.FC<Props> = ({
...
+  }, [count]);
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (onChange) {
+      onChange((prevMovie: Movie) => ({
+        ...prevMovie,
+        [name]: event.target.value,
+      }));
+    }
+  };
```
etojeDenys, line 45: better to move it to NewMovie component, it makes TextField not reusable ('cause of the Movie)

etojeDenys (approved):
good job

PR #2084, opened 2024-01-12
Develop

afesh4ak:
[DEMO LINK](https://afesh4ak.github.io/react_movies-list-add-form/)

there are some problems, I don't understand why 'error' is working correctly with title, and doesn't with other forms, also for 'Check if `imgUrl` and `imdbUrl` are valid URLs (you can use the next regex)

```js
const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;' pattern is too long,and I can't make it shorter.

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,108 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+interface Movie {
+  title: string;
+  description: string;
+  imgUrl: string;
+  imdbUrl: string;
+  imdbId: string;
+}
+
+interface NewMovieProps {
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: React.FC<NewMovieProps> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
   const [count] = useState(0);
+  const [formData, setFormData] = useState({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+
+  const [allFieldsFilled, setAllFieldsFilled] = useState(false);
+
+  useEffect(() => {
+    setAllFieldsFilled(
+      formData.title.trim() !== ''
+      && formData.imgUrl.trim() !== ''
+      && formData.imdbUrl.trim() !== ''
+      && formData.imdbId.trim() !== '',
+    );
+  }, [formData]);
+
+  const handleChange = (name: string, value: string) => {
+    setFormData({ ...formData, [name]: value });
+
+    setAllFieldsFilled(
+      formData.title.trim() !== ''
+      && formData.imgUrl.trim() !== ''
+      && formData.imdbUrl.trim() !== ''
+      && formData.imdbId.trim() !== '',
+    );
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (allFieldsFilled) {
+      onAdd(formData);
+
+      setFormData({
+        title: '',
+        description: '',
+        imgUrl: '',
+        imdbUrl: '',
+        imdbId: '',
+      });
+
+      setAllFieldsFilled(false);
+    }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/components/NewMovie/NewMovie.tsx [hunk 1]
```diff
@@ -1,45 +1,100 @@
...
+export const NewMovie: React.FC<NewMovieProps> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
etojeDenys, line 23: we can combine it in one useState
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,113 @@
...
+  const [allFieldsFilled, setAllFieldsFilled] = useState(false);
+
+  useEffect(() => {
+    setAllFieldsFilled(
+      formData.title.trim() !== ''
+      && formData.imgUrl.trim() !== ''
+      && formData.imdbUrl.trim() !== ''
+      && formData.imdbId.trim() !== '',
+    );
+  }, [formData]);
```
VitaliyBondarenko1982, line 36:
```suggestion

      const allFieldsFilled = formData.title.trim()
      && formData.imgUrl.trim()
      && formData.imdbUrl.trim()
      && formData.imdbId.trim();
```
it is redundant to manage state for allFieldsFilled. enough to write in constant. and expression can be shorter.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,113 @@
...
+
+  const handleChange = (name: string, value: string) => {
+    setFormData({ ...formData, [name]: value });
+
+    setAllFieldsFilled(
+      formData.title.trim() !== ''
+      && formData.imgUrl.trim() !== ''
+      && formData.imdbUrl.trim() !== ''
+      && formData.imdbId.trim() !== '',
+    );
```
VitaliyBondarenko1982, line 46:
```suggestion
```
it is looks like redundant. after change formData allFiedlsFilled will be recount.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,113 @@
...
+
+      setFormData({
+        title: '',
+        description: '',
+        imgUrl: '',
+        imdbUrl: '',
+        imdbId: '',
+      });
+
+      setAllFieldsFilled(false);
```
VitaliyBondarenko1982, line 63:
```suggestion
```
after change count form will be reinitialized and reset all prev states.

src/components/TextField/TextField.tsx
```diff
@@ -22,14 +23,20 @@ export const TextField: React.FC<Props> = ({
...
-  onChange = () => {},
+  onChange = () => { },
+  validate = () => true,
 }) => {
   // generage a unique id once on component load
   const [id] = useState(() => `${name}-${getRandomDigits()}`);
 
   // To show errors only if the field was touched (onBlur)
   const [touched, setTouched] = useState(false);
   const hasError = touched && required && !value;
```
VitaliyBondarenko1982, line 34:
```suggestion
  const hasError = touched && required && !value.trim();
```
it will fix such problem as in screen. I can fill some required inputs with only spaces and no see any errror, but still can't to add new movie.

chrome-capture-2024-1-13 (2)

VitaliyBondarenko1982 (changes requested):
Almost done.
Please check comments.

etojeDenys (approved):
lgtm

PR #2086, opened 2024-01-13
add task solution

vk-workshop:
[DEMO LINK](https://vk-workshop.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,106 @@
-import { useState } from 'react';
+/* eslint-disable max-len */
+import React, { useEffect, useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const [disabledBtn, setDisabledBtn] = useState(true);
+
+  const validateUrl = (url: string) => {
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,106 @@
...
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
VitaliyBondarenko1982, line 16:
```suggestion
  const [newMovie, setNewMovie] = useState({
     title: '',
     description: '',
     ...
  });
```
make one state for all form fields. use event.target.name for change proper field. pay attention that input should have attribute name should be the same as appropriate newMovie field.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,106 @@
...
+    setImdbId('');
+  };
+
+  useEffect(() => {
+    if (title && imgUrl && imdbUrl && imdbId) {
+      setDisabledBtn(false);
+    } else {
+      setDisabledBtn(true);
+    }
+  }, [title, imgUrl, imdbUrl, imdbId]);
```
VitaliyBondarenko1982, line 40:
```suggestion
const disabledBtn = !title.trim() || !imgUrl.trim() || !imdbUrl.trim() || !imdbId.trim()
```
it is redundant to manage state for such case. enough to make such check. trim every form field value should solve problem, when possible to add film with only spaces in each input (see screens)
Screenshot 2024-01-13 at 17 45 14
chrome-capture-2024-1-13 (1)

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comments.

vk-workshop:
- maked one state const [newMovie, setNewMovie]
- added disabledBtn for trim checking
- added validate implementation for imgUrl and imdbUrl

anastasiiavorobiova (changes requested):
Great work! Consider showing validation errors if the trimmed value is empty 
Screenshot 2024-01-14 at 16 04 00

vk-workshop:
- added checking for trimmed value when it is empty

anastasiiavorobiova (approved):
Great work!

PR #2091, opened 2024-01-15
solution

InnaSh23:
[DEMO LINK](https://InnaSh23.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,115 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
+import cn from 'classnames';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+// eslint-disable-next-line max-len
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,53 +1,126 @@
...
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    };
+
+    onAdd(newMovie);
+    setCount(count + 1);
+
+    // reset();
```
SanyaBratashchuk, line 52: remove commented code

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,53 +1,126 @@
...
+  };
+
+  useEffect(() => {
+    const isImgUrlValid = pattern.test(imgUrl);
+    const isImdbUrlValid = pattern.test(imdbUrl);
+
+    setIsFormValid(!!title && isImgUrlValid && isImdbUrlValid && !!imdbId);
+  }, [title, imgUrl, imdbUrl, imdbId]);
```
SanyaBratashchuk, line 60: There is no need in that useEffect. isFormValid can be just a common variable. 
```suggestion
  const isFormValid = !!title && pattern.test(imgUrl) && pattern.test(imdbUrl) && !!imdbId
```

and then in JSX
`<button ... disabled={isFormValid}>`

It should work in the same way

SanyaBratashchuk (changes requested):

maxim2310 (changes requested):
Good work, but lets make some improvement 

i try to add movie and fill in fields but button still disabled and haven't any error, looks strange 
image

SanyaBratashchuk (approved):
[Optional] would be better to add error like 'url is not valid` instead of `url is requied` in such cases 
image

PR #2096, opened 2024-01-16
solution

vSainiuk:
[DEMO LINK](https://vSainiuk.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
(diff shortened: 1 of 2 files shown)


DanilWeda (changes requested):
Perfect. 
I think you can upgrade your solution for the best code view! 
And if you want add validation for links) 
Many thanks!

olya-shyian (changes requested):
Great!💫

Try to fix this bug. Add some space to each input and then you will see that the 'add' button is no longer disabled, so we can create an empty movie and add it to the list. It should not work like that. Try to fix it

image

vSainiuk:
> Great!💫
> 
> Try to fix this bug. Add some space to each input and then you will see that the 'add' button is no longer disabled, so we can create an empty movie and add it to the list. It should not work like that. Try to fix it
> 
> image

I forgot to update link... I'm sorry)

olya-shyian (approved):
Cool!🤗

PR #2097, opened 2024-01-16
add solution

nabokok:
[DEMO LINK](https://nabokok.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
-import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+import './App.scss';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  const addMovie = (newMovie: Movie) => {
+    setMovies(prev => [...prev, newMovie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addMovie} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,92 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+const initialFields: Movie = {
+  title: '',
+  description: '',
+  imgUrl: '',
+  imdbUrl: '',
+  imdbId: '',
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [movieFields, setMovieFields] = useState<Movie>(initialFields);
+
+  const isButtonDisabled = Object.entries(movieFields)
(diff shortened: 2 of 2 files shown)


src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+import './App.scss';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  const addMovie = (newMovie: Movie) => {
+    setMovies(prev => [...prev, newMovie]);
+  };
```
artem8746, line 13: i think you don't need to create function for 1 line of code

artem8746 (approved):
Incredible job!!

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,92 @@
...
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    onAdd({
+      title: movieFields.title,
+      description: movieFields.description,
+      imgUrl: movieFields.imgUrl,
+      imdbUrl: movieFields.imdbUrl,
+      imdbId: movieFields.imdbId,
+    });
```
olya-shyian, line 39: you can pass just `movieFields`

onAdd(movieFields);

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,92 @@
...
+  const [count, setCount] = useState(0);
+  const [movieFields, setMovieFields] = useState<Movie>(initialFields);
+
+  const isButtonDisabled = Object.entries(movieFields)
+    .map(([key, value]) => (!value.trim() && key !== 'description'))
+    .some(Boolean);
+
+  const handleField = (name: string, value: string) => {
+    setMovieFields(prev => ({ ...prev, [name]: value }));
+  };
```
olya-shyian, line 29: this handler is great decision🔥

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,92 @@
...
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={movieFields.title}
+        onChange={(value) => handleField('title', value)}
```
olya-shyian, line 57: I'm not sure how does it work right now... but I think its better to write like this

` onChange={(event) => handleField('title', event.target.value)}`

check everywhere

olya-shyian (changes requested):
Great!🌟 
Just minor fixes needed

src/components/NewMovie/NewMovie.tsx
nabokok: We get event.target.value from TextField component.  It will not work if I write like this 
` onChange={(event) => handleField('title', event.target.value)}`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,86 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [movieFields, setMovieFields] = useState<Movie>(initialFields);
+
+  const isButtonDisabled = Object.entries(movieFields)
+    .map(([key, value]) => (!value.trim() && key !== 'description'))
+    .some(Boolean);
```
SerhiiKirik, line 25: easy operation would be with filter & check lenght

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,86 @@
...
+        value={movieFields.description}
+        onChange={(value) => handleField('description', value)}
       />
 
       <TextField
         name="imgUrl"
         label="Image URL"
-        value=""
+        value={movieFields.imgUrl}
+        onChange={(value) => handleField('imgUrl', value)}
```
SerhiiKirik, line 66: you can replace this
image
to =>
```js
  onChange={event => onChange(event)}
```
add `name` attribute to the `input`
after that in `onChangeCallback` take this name, `title` for example, from the event

SerhiiKirik (approved):
Well done 🔥

PR #2100, opened 2024-01-16
add solution

ssaviak:
[DEMO LINK](https://ssaviak.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -42,13 +42,14 @@ export const NewMovie: React.FC<Props> = ({ onAdd }: Props) => {
     setImgUrl('');
     setImdbUrl('');
     setImdbId('');
-    setCount(prevCount => prevCount + 1);
   };
 
-  const submitMovie = () => {
+  const submitMovie = (event: React.FormEvent<HTMLFormElement>) => {
     const movie = makeMovie();
 
     handleReset();
+    event.preventDefault();
+    setCount(prevCount => prevCount + 1);
 
     return onAdd(movie);
   };
@@ -57,10 +58,7 @@ export const NewMovie: React.FC<Props> = ({ onAdd }: Props) => {
     <form
       className="NewMovie"
       key={count}
-      onSubmit={(event) => {
-        event.preventDefault();
-        submitMovie();
-      }}
+      onSubmit={submitMovie}
     >
       <h2 className="title">Add a movie</h2>
 
@@ -83,7 +81,7 @@ export const NewMovie: React.FC<Props> = ({ onAdd }: Props) => {
         name="imgUrl"
         label="Image URL"
         value={imgUrl}
-        isValid={isValid(imgUrl)}
+        hasUrlError={!isValid(imgUrl)}
         onChange={(event) => setImgUrl(event)}
         required
       />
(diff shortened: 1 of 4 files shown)


HeySleare (approved):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,108 @@
...
+  };
 
   return (
-    <form className="NewMovie" key={count}>
+    <form
+      className="NewMovie"
+      key={count}
+      onSubmit={(event) => {
+        event.preventDefault();
+        submitMovie();
```
olya-shyian, line 62: I would move this `  event.preventDefault();` to `submitMovie`

```
const submitMovie = (event: React.FormEvent<HTMLFormElement>) => {
    event.preventDefault();
     ....
    onAdd(movieFields);
```
    
and here just write:
```onSubmit={handleSubmit}```

File: src/components/NewMovie/NewMovie.tsx [hunk 2]
```diff
@@ -1,45 +1,108 @@
...
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }: Props) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
olya-shyian, line 18: instead of creating a lot of states, use just one (recommendation for future)

```
const defaultMovie = {
    title: '',
    description: '',
    imgUrl: '',
    imdbUrl: '',
    imdbId: '',
  };

  const [newMovie, setNewMovie] = useState(defaultMovie);
```

currently its not necessary to change, you can leave like you implemented

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,108 @@
...
+    };
+  };
+
+  const handleReset = () => {
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+    setCount(prevCount => prevCount + 1);
```
olya-shyian, line 45: I would move this `setCount` to `submitMovie` function

src/components/TextField/TextField.tsx
```diff
@@ -23,13 +24,12 @@ export const TextField: React.FC<Props> = ({
...
+  isValid,
 }) => {
-  // generage a unique id once on component load
   const [id] = useState(() => `${name}-${getRandomDigits()}`);
-
-  // To show errors only if the field was touched (onBlur)
   const [touched, setTouched] = useState(false);
-  const hasError = touched && required && !value;
+  const hasError = touched && required && !value
+    && (isValid === undefined ? true : !isValid);
```
olya-shyian, line 32: its difficult to understand....could you please change the logic?

olya-shyian (changes requested):
Good work!🌟

It should show an error in this case that this url is not valid
image

And if I write like this I can not understand why button is still disabled
image

olya-shyian (approved):
You did a great job! Well done🤗

PR #2102, opened 2024-01-16
Solution

YevhenOstrovskiy:
[DEMO LINK](https://YevhenOstrovskiy.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -25,9 +25,23 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
 
   const handleAddForm = (event: React.FormEvent) => {
     event.preventDefault();
+    onAdd(
+      {
+        title,
+        description,
+        imgUrl,
+        imdbUrl,
+        imdbId,
+      },
+    );
+    clearForm();
   };
 
-  const isDisabled = !(title && imgUrl && imdbUrl && imdbId);
+
+  const isDisabled = !(title.trim()
+    && imgUrl.trim()
+    && imdbUrl.trim()
+    && imdbId.trim());
 
   return (
     <form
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -48,6 +83,19 @@ export const NewMovie = () => {
...
+                {
+                  title,
+                  description,
+                  imgUrl,
+                  imdbUrl,
+                  imdbId,
+                },
+              );
+              clearForm();
+            }}
```
sasholk, line 98: better make handler if callback more than 1 row

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,80 @@
...
+    setImdbUrl('');
+    setImdbId('');
+    setCount(() => count + 1);
+  }
+
+  const handleAddForm = (event: React.FormEvent) => {
+    event.preventDefault();
+  };
+
+  const isDisabled = !(title && imgUrl && imdbUrl && imdbId);
```
sasholk, line 30: add method `trim` cuz your form works with only spaces

sasholk (approved):
Nice Job Mate!! 🔥🔥🔥🔥
But make some changes

src/components/NewMovie/NewMovie.tsx
olya-shyian, line 98: I would recommend put all this logic to `handleAddForm` function

olya-shyian (changes requested):
Good work!💪

Unfortunately, there is a bug. If you add some space to each input then the 'add' button will no longer be disabled so you can create an empty movie and add to the list.

image

Please, try to fix it

olya-shyian (approved):
Nice🌸

PR #2106, opened 2024-01-16
completed add form

vpashko:
[DEMO LINK](https://vpashko.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,97 @@
-import { useState } from 'react';
+import React from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+const InitialFieldsOfMovie: Movie = {
+  title: '',
+  description: '',
+  imgUrl: '',
+  imdbUrl: '',
+  imdbId: '',
+};
+
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
+type Field =
+'title'
+| 'description'
+| 'imgUrl'
+| 'imdbUrl'
+| 'imdbId';
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = React.useState(0);
+  const [fieldsOfMovie, setFieldsOfMovie] = React
+    .useState(InitialFieldsOfMovie);
+
+  const isDisabled = !(
+    fieldsOfMovie.title
+    && fieldsOfMovie.imgUrl
+    && fieldsOfMovie.imdbUrl
+    && fieldsOfMovie.imdbId);
+
+  const submitHandler = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    onAdd({
+      title: fieldsOfMovie.title,
+      description: fieldsOfMovie.description,
+      imgUrl: fieldsOfMovie.imgUrl,
+      imdbUrl: fieldsOfMovie.imdbUrl,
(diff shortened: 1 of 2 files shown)


VasiliyYarosh (approved):
God job!

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,97 @@
-import { useState } from 'react';
+import React from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+const InitialFieldsOfMovie: Movie = {
```
SerhiiKirik, line 5:
```suggestion
const initialFieldsOfMovie: Movie = {
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,97 @@
...
+  imgUrl: '',
+  imdbUrl: '',
+  imdbId: '',
+};
+
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
+type Field =
```
SerhiiKirik, line 17: do you use this type?

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,97 @@
...
+'title'
+| 'description'
+| 'imgUrl'
+| 'imdbUrl'
+| 'imdbId';
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = React.useState(0);
+  const [fieldsOfMovie, setFieldsOfMovie] = React
+    .useState(InitialFieldsOfMovie);
```
SerhiiKirik, line 27: Just import `useState` from `'react'`
```suggestion
  const [fieldsOfMovie, setFieldsOfMovie] = useState(InitialFieldsOfMovie);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,97 @@
...
+    fieldsOfMovie.title
+    && fieldsOfMovie.imgUrl
+    && fieldsOfMovie.imdbUrl
+    && fieldsOfMovie.imdbId);
+
+  const submitHandler = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    onAdd({
+      title: fieldsOfMovie.title,
```
SerhiiKirik, line 39: would be better to create variable for it and after that put to `onAdd`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,97 @@
...
+    <form className="NewMovie" key={count} onSubmit={submitHandler}>
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={fieldsOfMovie.title}
+        onChange={(value) => changeFieldHandler('title', value)}
```
SerhiiKirik, line 62: you can replace this
image
to =>
```js
  onChange={event => onChange(event)}
```
add `name` attribute to the `input`
after that create `changeFieldHandler ` function, in this take this name, `title` for example, from the event

SerhiiKirik (changes requested):
Great 🔥 
Let's fix some problems 🧑‍💻 
image
image

src/components/NewMovie/NewMovie.tsx
vpashko: I use this type to check if I pass a valid value during an onChange event, but maybe it's not necessary, I'll fix it.

YegorKorenkov (approved):
Well done!

PR #2109, opened 2024-01-16
add movie-list form validation

sasholk:
[DEMO LINK](https://sasholk.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,22 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState([...moviesFromServer]);
+
+  const handleOnAdd = (movie: Movie) => setMovies([...movies, movie]);
+
   return (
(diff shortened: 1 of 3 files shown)


EdenParadise (approved):
I have nothing to add, great job, good code!

src/App.tsx
```diff
@@ -1,16 +1,27 @@
...
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
+
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={handleOnAdd} key={date} />
```
olya-shyian, line 24: `NewMovie` is not part of a dynamically generated list, the key prop may not be necessary in this specific case.

olya-shyian (changes requested):
Great work! 💪

If you create date only for `key` and don't use it anywhere except the `key` then I would remove it

src/App.tsx
sasholk: Maybe, I thought as you said before, but after the morning lesson with Misha he showed that solution and I like it. Thought that better idea than my previous 
image

olya-shyian (approved):
I got it! Thanks for the response🙂

PR #2116, opened 2024-01-18
New movie

KatiaGamal:
[DEMO LINK](https://KatiaGamal.github.io/react_movies-list-add-form/)


etojeDenys (changes requested):
i can create the movie only with spaces
image
let's fix that

TarasHoliuk (approved):
Nicely done!

PR #2118, opened 2024-01-18
add task solution

dyarechuk:
[DEMO LINK](https://dyarechuk.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,87 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+// eslint-disable-next-line max-len
+const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
+
+type Props = {
+  onAdd: (movie: Movie) => void,
+};
+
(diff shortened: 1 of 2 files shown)


avakiel (approved):
Nice job!

olya-shyian (changes requested):
Good work!
Try to fix one bug, inputs with url don't show error massage when I write not valid url...and as a result a user will not understand why 'add' button is still disabled

image

dyarechuk:
> Good work!
> Try to fix one bug, inputs with url don't show error massage when I write not valid url...and as a result a user will not understand why 'add' button is still disabled
> 
> image
> 

So, if url is not valid, this textfield has to highlight with red color and show a title, right?

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,89 @@
...
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+// eslint-disable-next-line max-len
+const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
+
+type Props = {
```
DanilWeda, line 8:
```suggestion
interface Props = {
```
use interface prefer

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,89 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isUrlValid = (url: string) => {
+    return pattern.test(url);
+  };
```
DanilWeda, line 21: you can move it outside component file

DanilWeda (approved):
Good job! 
Approved with small comments. 
Many thanks!

PR #2121, opened 2024-01-20
add task solution

yepolotn1ak:
[DEMO LINK](https://yepolotn1ak.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,112 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   const [count] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [submitDisabled, setSubmitDisabled] = useState(true);
+
+  const handleTitleChange = (value: string) => {
+    setTitle(value);
+    setSubmitDisabled(
+      !value.trim() || !value.trim() || !imdbUrl.trim() || !imdbId.trim(),
+    );
+  };
+
+  const handleImgUrlChange = (value: string) => {
+    setImgUrl(value);
+    setSubmitDisabled(
+      !title.trim() || !value.trim() || !imdbUrl.trim() || !imdbId.trim(),
+    );
+  };
+
+  const handleImdbUrlChange = (value: string) => {
+    setImdbUrl(value);
+    setSubmitDisabled(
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,112 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   const [count] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
etojeDenys, line 15:
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -71,42 +67,74 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
...
         label="Title"
         value={title}
-        onChange={handleTitleChange}
+        onChange={(event) => {
+          handleChange(event);
+          setSubmitDisabled(
+            !event.target.value.trim()
+            || !imgUrl.trim()
+            || !imdbUrl.trim()
+            || !imdbId.trim(),
```
yepolotn1ak, line 80: I can figure out how to call this method in the handleChange function and have it automatically check the input whose value we are changing and all the others.

I tried to write like this:

  setSubmitDisabled(!title.trim() || !imgUrl.trim() || !imdbUrl.trim() || !imdbId.trim()), but in this case, if for example I change the title, then this method analyzes the old state title, not current

anastasiiavorobiova (changes requested):
Feel free to ask for help in the chat 
Screenshot 2024-01-21 at 15 23 16
Consider showing errors if the input is touched and there are just spaces

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,139 @@
...
+        value={imgUrl}
+        onChange={(event) => {
+          handleChange(event);
+          setSubmitDisabled(
+            !title.trim()
+            || !event.target.value.trim()
+            || !imdbUrl.trim()
+            || !imdbId.trim(),
+          );
+        }}
```
volodymyr-soltys97, line 104: This logic is now duplicated in several places, you need to move it to a helper function and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,120 @@
...
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [submitDisabled, setSubmitDisabled] = useState(true);
```
etojeDenys, line 11: you don't need the state for that, it can be the simple var

etojeDenys (changes requested):
feel free to ask for some help with that in the chat

src/components/NewMovie/NewMovie.tsx
yepolotn1ak: for count ?

VitaliyBondarenko1982 (approved):
Looks good.

PR #2122, opened 2024-01-21
Develop

AV1788:
[DEMO LINK](https://AV1788.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,61 +1,133 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
-  const [text, setText] = useState('');
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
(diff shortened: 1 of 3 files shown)


anastasiiavorobiova (changes requested):
Great work! 
A user can create a film with spaces only, consider fixing it

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,56 +1,139 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
Viktor-Kost, line 17: It's not advisable to generate such a large number of states when we only have one instance (movie). 

```
const initialMovieState = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: ''
}
```

Example: 

```
}
```

```suggestion
  const [movie, setMovie] = useState(initialMovieState);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,56 +1,139 @@
...
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
+
+  const handleSubmit = (event:React.FormEvent) => {
+    event.preventDefault();
+    const trimmedTitle = title.trim()
```
Viktor-Kost, line 49: still can send a form with spaces in fields 
Screenshot 2024-01-22 at 23 08 35

Viktor-Kost (changes requested):

etojeDenys (approved):
lgtm

PR #2133, opened 2024-01-30
solution

Dmytrosp:
[DEMO LINK](https://Dmytrosp.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,10 +1,47 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+const EMPTY = '';
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState(EMPTY);
+  const [description, setDescription] = useState(EMPTY);
+  const [imgUrl, setImgUrl] = useState(EMPTY);
+  const [imdbUrl, setImdbUrl] = useState(EMPTY);
+  const [imdbId, setImdbId] = useState(EMPTY);
+
+  const activeButton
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,47 @@
...
+const EMPTY = '';
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState(EMPTY);
+  const [description, setDescription] = useState(EMPTY);
+  const [imgUrl, setImgUrl] = useState(EMPTY);
+  const [imdbUrl, setImdbUrl] = useState(EMPTY);
+  const [imdbId, setImdbId] = useState(EMPTY);
```
Viktor-Kost, line 18: It's not advisable to generate such a large number of states when we only have one instance (movie). 

```
const initialMovieState = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: ''
}
```

Example: 

```
}
```

```suggestion
  const [movie, setMovie] = useState(initialMovieState);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,47 @@
...
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+const EMPTY = '';
```
Viktor-Kost, line 9: redundant variable

src/components/TextField/TextField.tsx
```diff
@@ -24,12 +27,17 @@ export const TextField: React.FC<Props> = ({
...
-  // To show errors only if the field was touched (onBlur)
   const [touched, setTouched] = useState(false);
   const hasError = touched && required && !value;
+  let hasValidationError = false;
+
+  if (value && (name === 'imgUrl' || name === 'imdbUrl')) {
+    if (!value.match(pattern)) {
+      hasValidationError = true;
+    }
+  }
```
Viktor-Kost, line 40: it's better to move this logic to a helper function outside the component

Viktor-Kost (changes requested):
Also please check your deployment since DEMO link leads to 404 page now

xiimiik (approved):
lgtm

PR #2143, opened 2024-02-05
add task solution

Danylo-Chumakov:
[DEMO LINK](https://Danylo-Chumakov.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -26,9 +26,14 @@ export const NewMovie: React.FC<NewMovieProps> = ({ onAdd }) => {
   const handleBlur = (name: keyof typeof formData) => {
     if (name === 'titles' || name === 'imgUrl' || name === 'imdbUrl'
     || name === 'imdbId') {
+      setFormData((prevData) => ({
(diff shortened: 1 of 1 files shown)


IvanFesenko (changes requested):
Good start - check my comments:
- Add button should be disabled in case form not valid
- In case i type all spaces - there are no error messsages
2024-02-05_15h38_17
- After adding all fields are show error, but shouldn't

Danylo-Chumakov:
I corrected the first two points but I can’t solve the third. Can you point out the error?

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,142 @@
...
+
+    const newMovie: Movie = {
+      title: formData.titles,
+      description: formData.description,
+      imgUrl: formData.imgUrl,
+      imdbUrl: formData.imdbUrl,
+      imdbId: formData.imdbId,
+    };
+
+    onAdd(newMovie);
```
etojeDenys, line 72:
```suggestion
    onAdd(formData);
```

etojeDenys (approved):
good job

PR #2148, opened 2024-02-10
add movie to movies list

mixelio:
[DEMO LINK](https://mixelio.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,150 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,147 @@
...
+  const [description, setDescription] = useState('');
+
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbUrlError, setiMdbUrlError] = useState(false);
+
+  const [imgUrl, setImgUrl] = useState('');
+  const [imgUrlError, setImgUrlError] = useState(false);
+
+  const [imdbId, setImdbId] = useState('');
+  const [imdbIdError, setImdbIdError] = useState(false);
```
etojeDenys, line 25: we can combine it in one useState
```
  title: '',
  ...
})
```

```
}
```

etojeDenys: you can do the same with errors

etojeDenys (changes requested):

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
There is now an option to add empty fields, this needs to be fixed
image
image

etojeDenys (approved):
good job

PR #2151, opened 2024-02-13
Movies list add form

roman-tarasov90:
https://roman-tarasov90.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,85 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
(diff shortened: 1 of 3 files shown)


etojeDenys (changes requested):
we can combine it in one useState
```
  title: '',
  ...
})
```

```
}
```

roman-tarasov90:
> we can combine it in one useState
> 
> ```
> const [newMovie, setNewMovie] = useState({
>   title: '',
>   ...
> })
> ```
> 
> now we can also create only one handle function for all fields
> 
> ```
> const handleChange = (e: ...) => {
>   const {name, value} = e.target;
>   setMovie((prevMovie) => ({...prevMovie, [name]: value}))
> }
> ```
> 
> don't forget to specify name attr to input in TextField

I can solve it only with chat GPT. It's not a simplificatio in that lvl in my opinion. So:
1. Why i should use [name], value?
2. What's 'e'? I can't get any event there. As a result I break everything

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,87 @@
...
+      imdbUrl: '',
+      imdbId: '',
+    });
+  };
+
+  const isFilled
+    = newMovie.title.trim().length > 0
+    && newMovie.imgUrl.trim().length > 0
+    && newMovie.imdbUrl.trim().length > 0
+    && newMovie.imdbId.trim().length > 0;
```
Andriy-Shymkiv, line 42:
```suggestion
    = newMovie.title.trim().length
    && newMovie.imgUrl.trim().length
    && newMovie.imdbUrl.trim().length
    && newMovie.imdbId.trim().length;
```

Andriy-Shymkiv (approved):
good job!

PR #2152, opened 2024-02-14
Solution

artemvlasiuk:
[DEMO LINK](https://artemvlasiuk.github.io/react_movies-list-add-form/)


etojeDenys (changes requested):

anastasiiavorobiova (changes requested):
Consider showing errors if there are just spaces in the input

etojeDenys (approved):
lgtm

PR #2158, opened 2024-02-17
solution

isaeev:
[DEMO LINK](https://isaeev.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,45 +1,125 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  'onAdd': (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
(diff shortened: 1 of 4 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The use cannot be able to add the empty fields, fix it
image
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,125 @@
...
+  const [descriptionInput, setDescriptionInput] = useState('');
+
+  const [imgUrlInput, setImgUrlInput] = useState('');
+  const [hasImgUrlInputError, setHasImgUrlInputError] = useState(true);
+
+  const [imdbUrlInput, setImdbUrlInput] = useState('');
+  const [hasImdbUrlInputError, setHasImdbUrlInputError] = useState(true);
+
+  const [imdbIdInput, setImdbIdInput] = useState('');
+  const [hasImdbIdInputError, setImdbIdInputError] = useState(true);
```
etojeDenys, line 26: we can combine it in one useState
```
  title: '',
  ...
})
```

```
}
```

!! don't forget to specify name attr to input in TextField

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #2161, opened 2024-02-19
Develop

kostyaniekrasov:
[DEMO LINK](https://kostyaniekrasov.github.io/react_movies-list-add-form/)
for some reason, when I use the trim() method for btnDisabled, half of the tests fail.

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,86 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitleField] = useState('');
+  const [description, setDescriptionField] = useState('');
+  const [imgUrl, setImageUrlField] = useState('');
+  const [imdbUrl, setImdbUrlField] = useState('');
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,94 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitleField] = useState('');
+  const [description, setDescriptionField] = useState('');
+  const [imgUrl, setImageUrlField] = useState('');
+  const [imdbUrl, setImdbUrlField] = useState('');
+  const [imdbId, setImdbIdField] = useState('');
```

```
const initialMovieState = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: ''
}
```

4. Whenever you need to change value of any field you should pass `event` and then just change a value of a movie object by key. In this case you will reuse one handler instead of handlers for each input
Example: 

```
  const {name, value} = e.target; setMovie(prevInputs => ({...prevInputs, [name]: value}))
}
```

```suggestion
  const [movie, setMovie] = useState(initialMovieState);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,94 @@
...
+
+  const btnDisabled = !title || !imgUrl || !imdbUrl || !imdbId;
```
Viktor-Kost, line 17: yes, you should use `trim` here to avoid adding just `spaces`. 
If you have some problems with this implementation feel free to ask at fe_chat

Viktor-Kost (changes requested):

anastasiiavorobiova (changes requested):
Great work! Consider showing error messages when a user adds just spaces into your fields 
Screenshot 2024-02-21 at 16 42 45
If you feel that some of the comments are unclear, please feel free to ask for help in the chat or come to the Q&A session. We will deal with everything)

etojeDenys (approved):
lgtm

PR #2162, opened 2024-02-20
form implemented

Kyryb:
[DEMO LINK](https://kyryb.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -9,90 +9,39 @@ type Props = {
 export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
+  const CLEAN_FORM: Movie = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
   // eslint-disable-next-line max-len
   const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
-  const [movie, setMovie] = useState({
-    title: {
-      value: '',
-      isRequired: true,
-    },
-    description: {
-      value: '',
-      isRequired: false,
-    },
-    imgUrl: {
-      value: '',
-      isRequired: true,
-    },
-    imdbUrl: {
-      value: '',
-      isRequired: true,
-    },
-    imdbId: {
-      value: '',
-      isRequired: true,
-    },
-  });
+  const [movie, setMovie] = useState(CLEAN_FORM);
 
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,106 @@
...
+  const [title, setTitle] = useState('');
+  const isTitleRequired = true;
+  const [description, setDescription] = useState('');
+  const isDescriptionRequired = false;
+  const [imgUrl, setImgUrl] = useState('');
+  const isImageUrlRequired = true;
+  const [imdbUrl, setImdbUrl] = useState('');
+  const isImdbUrlRequired = true;
+  const [imdbId, setImdbId] = useState('');
+  const isImdbIdRequired = true;
```
etojeDenys, line 24: we can combine it in one useState
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):

Kyryb:
no idea if it's good approach to store isRequired in movie UseState :)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,148 @@
...
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
-        required
+        value={movie.title.value}
+        onChange={handleChange}
+        required={movie.title.isRequired}
```
TarasHoliuk, line 111: > no idea if it's a good approach to store isRequired flag in movie UseState :)

good point :slightly_smiling_face: - it's not
You never change it - so there is no reason to store it in state (no need to store it anywhere - just pass boolean prop to `TextField` if it's required or not - if it's not)

```suggestion
        required
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,45 +1,148 @@
...
+      ...prevMovie,
+      [name]: {
+        ...prevMovie[name as keyof typeof prevMovie],
+        value,
+      },
+    }));
+  };
+
+  const resetForm = () => {
+    setMovie(prevMovie => ({
```
TarasHoliuk, line 59: you may create an object with default values like this

```
const defaultValue = {
  title: '',
  // rest input values
};
```
above component and use it as default state value and in `resetForm` function

src/components/TextField/TextField.tsx
```diff
@@ -41,20 +44,24 @@ export const TextField: React.FC<Props> = ({
...
           className={classNames('input', {
-            'is-danger': hasError,
+            'is-danger': hasError || hasValidationError,
           })}
           placeholder={placeholder}
           value={value}
-          onChange={event => onChange(event.target.value)}
+          onChange={event => onChange(event)}
           onBlur={() => setTouched(true)}
+          // onFocus={() => setTouched(false)}
```
TarasHoliuk, line 56: remove comment

TarasHoliuk (changes requested):

src/components/NewMovie/NewMovie.tsx
Kyryb: That's what i did first but then i stuck trying to make my submit button disabled if any field has an error, because description is not required like other fields and i can't access it outside of component

TarasHoliuk (approved):
:heavy_check_mark:

PR #2171, opened 2024-02-26
Solution

Krzychu92:
https://krzychu92.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,99 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  addMovie: (newMovie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ addMovie }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [newMovie, setNewMovie] = useState({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+  let isDisable = true;
+
+  const reset = () =>
+    setNewMovie({
+      title: '',
+      description: '',
+      imgUrl: '',
+      imdbUrl: '',
+      imdbId: '',
+    });
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { name, value } = event.target;
+
+    setNewMovie(currentMovie => ({ ...currentMovie, [name]: value }));
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (isDisable) {
+      return;
+    }
+
+    addMovie(newMovie);
+
+    setCount(actualyCount => actualyCount + 1);
+    reset();
+  };
+
+  if (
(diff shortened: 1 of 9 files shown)


yevhenii-pyl (commented):
`NewMovie should check if title, imgUrl, imdbUrl, imdbId are entered when an input looses focus (onBlur) and show an error and a red border if needed (learn how it it implemented in the TextField);`

There is no validation when an input looses focus.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,107 @@
...
+type Props = {
+  addMovie: (newMovie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ addMovie }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [newMovie, setNewMovie] = useState({
```
yevhenii-pyl, line 13: I'd rather create a separate useState for each property, it will make validation way simpler.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,107 @@
...
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+  let isDisable = true;
```
yevhenii-pyl, line 20: Should be a state property named like `submitDisabled` or something.

src/components/NewMovie/NewMovie.tsx
yevhenii-pyl, line 13: I'd rather create a constant for `initialMovieState` outside the component .

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,107 @@
...
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { name, value } = event.target;
+
+    setNewMovie(currentMovie => ({ ...currentMovie, [name]: value }));
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (isDisable) {
```
yevhenii-pyl, line 40: No need to explicitly check it here since the button is disabled if `isDisable`.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,107 @@
...
+      return;
+    }
+
+    addMovie(newMovie);
+
+    setCount(actualyCount => actualyCount + 1);
+    reset();
+  };
+
+  if (
```
yevhenii-pyl, line 50: Very strange code, you can initially declare the `isDisable` value based on these conditions. 

like `let isDisable = !newMovie.title.trim() ...`

yevhenii-pyl (changes requested):
Overall, a good job. Let's just adjust a few things to make it better.

src/components/NewMovie/NewMovie.tsx
Krzychu92: I hope I have fixed all issues

yevhenii-pyl (approved):
There are no changes when I check your demo link. 
However, I see required changes in the last commit. I have cloned your repo and tried it locally - it works just fine so I see no sense in blocking your PR further. 

BUT!
I leave it on you to double-check your demo link and find out what exactly gone wrong with new version deploy. Maybe, it just needs to take it's time to update, it also happens.

Krzychu92:
The demo is now working fine ;)

PR #2178, opened 2024-03-01
solution

Yehorf21:
[DEMO LINK](https://Yehorf21.github.io/react_movies-list-add-form/)


l4st1m0za (changes requested):
Good job, but you need some changes:

1) User can add a movie even though, it's URL's are wrong
image
image

l4st1m0za (approved):
Good job!

PR #2187, opened 2024-03-06
Develop

wsadocha:
[DEMO LINK](https://wsadocha.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,131 @@
...
+      [name]: value,
+    });
+
+    setValidity({
+      ...validity,
+      [name]: validateInput(value),
+    });
+  };
+
+  const isValid = (): boolean => {
```
yevhenii-pyl, line 46: Can be just a variable, isn't it? :)

`const isValid = Object.values(validity).every(valid => valid);`

PR #2188, opened 2024-03-07
Add solution

Vasya564:
[DEMO LINK](https://Vasya564.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -15,13 +15,11 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   const [imdbUrl, setImdbUrl] = useState('');
   const [imdbId, setImdbId] = useState('');
 
-  const isAllFilled = () => {
-    if (!title.trim() || !imgUrl.trim() || !imdbUrl.trim() || !imdbId.trim()) {
-      return true;
-    }
-
-    return false;
-  };
+  // prettier-ignore
+  const isAllFieldsEmpty = !title.trim()
+    || !imgUrl.trim()
+    || !imdbUrl.trim()
+    || !imdbId.trim();
 
   const handleSubmit = (event: React.FormEvent) => {
     event.preventDefault();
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,97 @@
...
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isAllFilled = () => {
+    if (!title.trim() || !imgUrl.trim() || !imdbUrl.trim() || !imdbId.trim()) {
+      return true;
+    }
+
+    return false;
+  };
```
BudnikOleksii, line 24: - The name is `isAllFilled`, but if some of them are empty you return true 🤷 
- You already have boolean values from `!title.trim()` so you don't need `if` and then return true/false
- And actually you don't need a function, just a variable

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,97 @@
...
+    <form className="NewMovie" key={count} onSubmit={handleSubmit}>
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={value => setTitle(value)}
```
BudnikOleksii, line 53:
```suggestion
        onChange={setTitle}
```
Could be simplified, fix in all places

BudnikOleksii (changes requested):
Good job 👍 
Just a few fixes are required

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,95 @@
...
+      />
 
-      <TextField name="imdbId" label="Imdb ID" value="" />
+      <TextField
+        name="imdbId"
+        label="Imdb ID"
+        value={imdbId}
+        onChange={setImdbId}
+        required
+      />
```
stassokolov1, line 84: will be better in future iterate these `TextField` using `.map()` (not critical now)

StasSokolov1 (approved):
Looks good to me🔥

PR #2191, opened 2024-03-07
solution

yurovych:
https://yurovych.github.io/react_movies-list-add-form/


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,142 @@
...
+      imdbId: '',
+    });
+  };
+
+  function cheeckBeforeSubmit() {
+    if (Object.values(neccesarlyField).some(item => item.trim() === '')) {
+      return true;
+    }
+
+    return false;
```
SanyaBratashchuk, line 46:
```suggestion
    return Object.values(neccesarlyField).some(item => item.trim() === '');
```

SanyaBratashchuk (approved):

PR #2194, opened 2024-03-10
working solution

Patryk-Buczkowski:
[DEMO LINK](https://Patryk-Buczkowski.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [newMovie, setNewMovie] = useState(moviesFromServer);
+
+  const movieAdd = (movie: Movie) => {
+    setNewMovie(current => [...current, movie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={newMovie} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={movie => movieAdd(movie)} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,133 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie = ({ onAdd }: Props) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [newMovie, setNewMovie] = useState({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+
+  const handleTrim = () => {
+    if (
+      Object.values(newMovie)
+        .filter(key => key !== newMovie.description)
+        .some(item => item.trim() === '')
+    ) {
+      return true;
+    }
+
+    return false;
+  };
+
+  // eslint-disable-next-line max-len, prettier/prettier
+  const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
+
+  const handleTitleChange = (newVal: string) => {
+    setNewMovie(current => ({
+      ...current,
+      title: newVal,
+    }));
+  };
+
+  const handleDescriptionChange = (newVal: string) => {
+    setNewMovie(current => ({
+      ...current,
+      description: newVal,
+    }));
+  };
+
+  const handleImgUrlChange = (newVal: string) => {
+    if (pattern.test(newVal)) {
+      setNewMovie(current => ({
+        ...current,
+        imgUrl: newVal,
+      }));
+    }
+  };
+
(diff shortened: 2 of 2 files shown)


src/App.tsx
```diff
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [newMovie, setNewMovie] = useState(moviesFromServer);
```
yevhenii-pyl, line 9: Something is odd with the naming here. 

You call it '`newMovie`, but pass `moviesFromServer` as an initial state? 
Probably, you want to call it like, IDK, `movies`?

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,142 @@
...
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie = ({ onAdd }: Props) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [newMovie, setNewMovie] = useState({
```
yevhenii-pyl, line 13: This is more likely to be called `newMovie` :)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,142 @@
...
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+
+  const handleTrim = () => {
```
yevhenii-pyl, line 21: It looks like some sort of a validator function. Something with the naming once again. 

You return `true` if any key, except of one for description (which is not a required field), is empty. That makes sense, but you probably should consider some other naming.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,142 @@
...
+    }
+
+    return false;
+  };
+
+  const pattern =
+    // eslint-disable-next-line max-len
+    /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
+
+  const handleTitleChange = (newVal: string) => {
```
yevhenii-pyl, line 37: Let's see: you have 5 similar functions here. Can't it be just one function that will receive the key and value?

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,142 @@
...
+        .filter(key => key !== newMovie.description)
+        .some(item => item.trim() === '')
+    ) {
+      return true;
+    }
+
+    return false;
+  };
+
+  const pattern =
```
yevhenii-pyl, line 33: Another naming :) 
Call it in a more explicit manner, like what exactly this pattern is validating?

yevhenii-pyl (changes requested):
That's a progress, but let's refactor a bit 👍 
Just pay more attention to naming and your code will become way cleaner and more understandable for you at the first place ;)

src/App.tsx
```diff
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setNewMovie] = useState(moviesFromServer);
```
yevhenii-pyl, line 9: movies, setMovies :)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,106 @@
...
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie = ({ onAdd }: Props) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [movies, setMovies] = useState({
```
yevhenii-pyl, line 13: Nope, naming here was absolutely fine. Was my comment misleading? I just wanted to emphasise that calling a state variable newMovie here was a right decision.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,106 @@
...
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+
+  const handleValidateButtonDisabled = () => {
```
yevhenii-pyl, line 21: just `validate` :)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,106 @@
...
+    setMovies(current => ({
+      ...current,
+      [key]: newVal,
+    }));
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    handleValidateButtonDisabled();
+    onAdd(movies);
```
yevhenii-pyl, line 43: You are validating here, but movie is added no matter what was the result of your validation function, as a result, you have these errors in tests: 

```
    1) should not clear the form if title is empty

    2) should not call onAdd if title is empty

    3) should not call onAdd if imgUrl is empty

    4) should not call onAdd if imdbUrl is empty

    5) should not call onAdd if imdbId is empty
```

All logic you have after the validation should depend on a validation result.

yevhenii-pyl (changes requested):
Let's refactor one more time :) 
Don't forget to check tests results

src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={movi => movieAdd(movi)} />
```
BogdanMaliuta, line 21: You need to pass here only your func like: 
```suggestion
        <NewMovie onAdd={movieAdd} />
```

BogdanMaliuta (approved):
Gj!Fix only 1 thing

PR #2198, opened 2024-03-12
solved task

hulchak:
https://hulchak.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -7,48 +7,43 @@ type Props = {
   onAdd: (movie: Movie) => void;
 };
 
+const initialMovieState = {
+  title: '',
+  description: '',
+  imgUrl: '',
+  imdbUrl: '',
+  imdbId: '',
+};
+
 export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
+  const [movie, setMovie] = useState(initialMovieState);
   const [count, setCount] = useState(0);
-  const [title, setTitle] = useState('');
-  const [description, setDescription] = useState('');
-  const [imgUrl, setImgUrl] = useState('');
-  const [imdbUrl, setImdbUrl] = useState('');
-  const [imdbId, setImdbId] = useState('');
-  const [titleTouched, setTitleTouched] = useState(false);
-  const [imgUrlTouched, setImgUrlTouched] = useState(false);
-  const [imdbUrlTouched, setImdbUrlTouched] = useState(false);
-  const [imdbIdTouched, setImdbIdTouched] = useState(false);
 
   const reset = () => {
-    setTitle('');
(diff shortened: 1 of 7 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,124 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
Viktor-Kost, line 18: It's not advisable to generate such a large number of states when we only have one instance (movie). 

```
const initialMovieState = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: ''
}
```

4. Whenever you need to change value of any field you should pass `event` argument and then just change a value of a movie object by key. In this case you will reuse one handler instead of handlers for each input
Example: 

```
  const {name, value} = e.target; setMovie(prevInputs => ({...prevInputs, [name]: value}))
}
```

```suggestion
  const [movie, setMovie] = useState(initialMovieState);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,124 @@
...
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+    setTitleTouched(false);
+    setImgUrlTouched(false);
+    setImdbUrlTouched(false);
+    setImdbIdTouched(false);
+    setCount(c => c + 1);
+  };
```
Viktor-Kost, line 35: 5. for reset function you will just pass initialMovieState

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,124 @@
...
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [titleTouched, setTitleTouched] = useState(false);
+  const [imgUrlTouched, setImgUrlTouched] = useState(false);
+  const [imdbUrlTouched, setImdbUrlTouched] = useState(false);
+  const [imdbIdTouched, setImdbIdTouched] = useState(false);
```
Viktor-Kost, line 22: You can work with this logic similar to the logic with `movie` I've written above

Viktor-Kost (changes requested):

l4st1m0za (approved):
Good job!

PR #2199, opened 2024-03-12
add task solution

ihor-pavlov:
[DEMO LINK](https://ihor-pavlov.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,127 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,135 @@
...
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (movie: Movie) => void;
+}
+
+// функція створення нового допису
```
volodymyr-soltys97, line 9: Comments must be only on English if you need it

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better.
You need to use `trim()` for the links
image

When I want to add a new film, you need to show a message that a title is required
image

VitaliyBondarenko1982 (changes requested):
Almost done.
Trim values of required fields during validation.
Now no see any error when typed all emty spaces and made onBlur.
chrome-capture-2024-3-14 (1)

etojeDenys (approved):
lgtm

PR #2204, opened 2024-03-13
movies

shelyaa:
[DEMO LINK](https://shelyaa.github.io/react_movies-list-add-form/)


IvanFesenko (changes requested):
Lets show user that only space not valid input thats why button is disabled
2024-03-13_16h49_42

IvanFesenko (approved):
Nice

PR #2223, opened 2024-03-18
form added

anastasia1383:
[DEMO LINK](https://anastasia1383.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,111 @@
 import { useState } from 'react';
+
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+import { Movie } from '../../types/Movie';
+
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+type State = {
+  title: string;
+  description: string;
+  imgUrl: string;
+  imdbUrl: string;
+  imdbId: string;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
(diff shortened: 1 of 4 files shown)


vlesia (approved):
Good job!

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,119 @@
...
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [state, setState] = useState<State>({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
```
BudnikOleksii, line 27:
```suggestion
  const [newMovie, setNewMovie] = useState<State>(DEFAULT_MOVIE);
```
- Create a default movie object so you can pass it here and then when resetting values
- `state` isn't informative name, you can have a lot of them

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,119 @@
...
+    .map((key: string): boolean => {
+      if (key === 'description') {
+        return false;
+      }
+
+      const value = state[key as keyof State];
+
+      return value === '' || (typeof value === 'string' && value.trim() === '');
+    })
+    .every(key => key === false);
```
BudnikOleksii, line 58: - It's too complicated, simplify it. You can just check the values. `!title.trim()`, etc.

BudnikOleksii (changes requested):
Good work 👍 
But could be even better, let's improve your solution slightly

StasSokolov1 (approved):
Looks good to me🔥

Not critical but will be good to add URL validation
Снимок экрана 2024-03-19 в 11 18 42

PR #2225, opened 2024-03-18
add solution

vlesia:
[DEMO LINK](https://vlesia.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,10 +1,51 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type NewMovieProps = {
+  onAdd: (newMovie: Movie) => void;
+};
+
+export const NewMovie = ({ onAdd }: NewMovieProps) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const [isFormValid, setIsFormValid] = useState(false);
+
+  const handleFormValidity = () => {
+    setIsFormValid(
+      title.trim() !== '' &&
+        imgUrl.trim() !== '' &&
+        imdbUrl.trim() !== '' &&
+        imdbId.trim() !== '',
+    );
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    const newMovie: Movie = {
+      title: title,
+      description: description,
+      imgUrl: imgUrl,
+      imdbUrl: imdbUrl,
+      imdbId: imdbId,
+    };
+
+    onAdd(newMovie);
+    setCount(count + 1);
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+    setIsFormValid(false);
+  };
 
   return (
     <form className="NewMovie" key={count}>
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -13,25 +54,65 @@ export const NewMovie = () => {
...
+      />
 
       <div className="field is-grouped">
         <div className="control">
           <button
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!isFormValid}
+            onClick={handleSubmit}
```
yaGodya, line 115: Great job.

I have a couple of tips. 

`disabled={!isFormValid} `can be placed in `handleSubmit `

it is advisable to hang `handleSubmit `on the entire `form `block

src/components/NewMovie/NewMovie.tsx
```diff
@@ -13,25 +54,65 @@ export const NewMovie = () => {
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={(newValue: string) => {
+          setTitle(newValue);
+          handleFormValidity();
```
yaGodya, line 60: It is not necessary to pass validation to every form element.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -13,25 +54,65 @@ export const NewMovie = () => {
...
       />
 
-      <TextField name="description" label="Description" value="" />
+      <TextField
+        name="description"
+        label="Description"
+        value={description}
+        onChange={(newValue: string) => {
+          setDescription(newValue);
+          handleFormValidity();
```
yaGodya, line 71: also

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,51 @@
...
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const [isFormValid, setIsFormValid] = useState(false);
+
+  const handleFormValidity = () => {
```
yaGodya, line 21: It is not necessary to create a function, it is enough to create a variable. 

` const isDisabled =
    !title.trim() || !imgUrl.trim() || !imdbUrl.trim() || !imdbId.trim();
`

and pass the result of this variable to the submit button

src/components/NewMovie/NewMovie.tsx
```diff
@@ -13,25 +54,65 @@ export const NewMovie = () => {
...
+        required
+      />
 
       <div className="field is-grouped">
         <div className="control">
           <button
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!isFormValid}
```
yaGodya, line 114: pass a variable of 21 lines here

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,51 @@
...
+    const newMovie: Movie = {
+      title: title,
+      description: description,
+      imgUrl: imgUrl,
+      imdbUrl: imdbUrl,
+      imdbId: imdbId,
+    };
+
+    onAdd(newMovie);
+    setCount(count + 1);
```
yaGodya, line 41: for a more correct update of the counter state, it is preferable to use
 `setCount(currentCount => currentCount + 1);`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,51 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type NewMovieProps = {
+  onAdd: (newMovie: Movie) => void;
+};
+
+export const NewMovie = ({ onAdd }: NewMovieProps) => {
```
yaGodya, line 9: if you use` React.FC<Props>` - it will allow you not to specify in each component the type of data that is passed inside

yaGodya (approved):
Good job.👍

I pass it on to the mentor for further verification

src/components/NewMovie/NewMovie.tsx
BudnikOleksii: > if you use` React.FC<Props>` - it will allow you not to specify in each component the type of data that is passed inside

Good point

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,98 @@
...
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [count, setCount] = useState(0);
+
+  const isFormValid = Boolean(
+    title.trim() !== '' &&
+      imgUrl.trim() !== '' &&
+      imdbUrl.trim() !== '' &&
+      imdbId.trim() !== '',
```
BudnikOleksii, line 24:
```suggestion
    title.trim() &&
      imgUrl.trim() &&
      imdbUrl.trim() &&
      imdbId.trim(),
```
You don't need to compare with an empty string, just check if value exists
https://developer.mozilla.org/en-US/docs/Glossary/Truthy

BudnikOleksii (changes requested):
Looks good 👍 
Just two minor fixes are required

BudnikOleksii (approved):
Good work 👍

PR #2233, opened 2024-03-18
add task solution

Math8585:
[DEMO LINK](https://Math8585.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,88 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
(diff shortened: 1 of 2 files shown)


IRONH4MMER (changes requested):
Good work! But pay attention to the third requirement:

> Disable the submit button until all the required fields are filled (spaces should be trimmed);

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,99 @@
...
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isEnteredValue =
+    title.trim() !== '' &&
+    imgUrl.trim() !== '' &&
+    imdbUrl.trim() !== '' &&
+    imdbId.trim() !== '';
```
BudnikOleksii, line 21:
```suggestion
  const isValidForm =
    title.trim() &&
    imgUrl.trim() &&
    imdbUrl.trim() &&
    imdbId.trim();
```
Don't compare with an empty string, just check if a value exists. https://developer.mozilla.org/en-US/docs/Glossary/Truthy

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,99 @@
...
+    <form className="NewMovie" onSubmit={handleSubmit} key={count}>
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={newValue => setTitle(newValue)}
```
BudnikOleksii, line 55:
```suggestion
        onChange={setTitle}
```
You can pass setState function directly, fix it in all places

BudnikOleksii (changes requested):
Looks good 👍 
Just two minor fixes are required

StasSokolov1 (approved):
The code and demo look like a masterpiece🔥

PR #2241, opened 2024-03-19
add movies list add form

natalia-serogina:
https://natalia-serogina.github.io/react_movies-list-add-form/


witflash (approved):
Good job!

A small recommendation for the future.
It's better to create a single state as an object rather than create many separate states.
For example, you can create states for the form and errors:
```
const [formFields, setFormFields] = useState({
  title: '',
  description: '',
  ...
  }

const [errors, setErrors] = useState({
  title: '',
  description: '',
  ...
  }
```
and then change a particular object prop.

PR #2245, opened 2024-03-19
add task solution

dima91020:
- [DEMO LINK](https://dima91020.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -7,35 +7,39 @@ type Props = {
 };
 
 export const NewMovie: React.FC<Props> = ({ onAdd }) => {
-  const [title, setTitle] = useState('');
-  const [description, setDectiption] = useState('');
-  const [imgUrl, setImgUrl] = useState('');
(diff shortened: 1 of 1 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,97 @@
...
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [title, setTitle] = useState('');
+  const [description, setDectiption] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
loralevitska, line 14: Could be combined into 1 state: { title: "", description: "", .... }

loralevitska (changes requested):
Try to avoid adding empty fields 
Use [trimming](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) for fields
image

etojeDenys (approved):
lgtm

PR #2248, opened 2024-03-20
add solution

Dmitrel:
https://Dmitrel.github.io/react_movies-list-add-form/


Vasya564 (changes requested):
Good job🚀 
Almost done, after adding a new movie form should be cleared

PR #2251, opened 2024-03-20
Develop

Viktoria-Moroz:
[DEMO LINK](https://Viktoria-Moroz.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -6,7 +6,7 @@ type Props = {
   onAdd: (movie: Movie) => void;
 };
 
-export const NewMovie: React.FC<Props> = ({ onAdd: onSubmit }) => {
+export const NewMovie = ({ onAdd }: Props) => {
   // Increase the count after successful form submission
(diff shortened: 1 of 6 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,34 +1,123 @@
...
+  };
+
+  const handleImdbUrl = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setImdbUrl(event.target.value);
+  };
+
+  const handleImdbId = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setImdbId(event.target.value);
+  };
+  // #endregion
```
loralevitska, line 40: Remove comments. Also all handlers can be combined into 1

For this you should change the state structure:
state = {
  title: "",
  description: "",
  ......
}

const handleClick = (event: React.ChangeEvent<HTMLInputElement>) => {
    setState(prev => ({ ...prev, [name]: value }));
};

loralevitska (changes requested):
Good job! There are things that should be fixed:

PR #2261, opened 2024-03-26
solution with optional task

IhorRome:
[DEMO LINK](https://IhorRome.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,93 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
anastasiia-tilikina, line 17: let's combine these states into one state with object.
Also lets create variable initialValues and set it as initial state and then reuse when we clear form.

Then we can create one handler to change particular field in this object

src/components/TextField/TextField.tsx
```diff
@@ -27,7 +31,18 @@ export const TextField: React.FC<Props> = ({
 
   // To show errors only if the field was touched (onBlur)
   const [touched, setTouched] = useState(false);
-  const hasError = touched && required && !value;
+  let hasError = touched && required && !value;
+
+  let errorMessage = `${label} is required`;
```
anastasiia-tilikina, line 36: It would be nice to show some error message when field contains spaces only since it is not valid and submit is disabled.

anastasiia-tilikina (changes requested):
Well done!

let's just fix a couple of issues here

src/components/TextField/TextField.tsx
```diff
@@ -45,12 +67,12 @@ export const TextField: React.FC<Props> = ({
           })}
           placeholder={placeholder}
           value={value}
-          onChange={event => onChange(event.target.value)}
-          onBlur={() => setTouched(true)}
+          onChange={event => onChange(name as keyof Movie, event.target.value)}
```
Viktor-Kost, line 70:
```suggestion
          onChange={onChange}
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,141 @@
...
+  };
+
+  const handleOnChange = (field: keyof Movie, newValue: string): void => {
+    setMovie(prev => {
+      const newMovie = { ...prev };
+
+      newMovie[field] = newValue;
+
+      return newMovie;
+    });
```
Viktor-Kost, line 61:
```suggestion
  const { name, value } = e.target; 
  setMovie(prevInputs => ({...prevInputs, [name]: value}))

```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,141 @@
...
+    setMovieError(prev => {
+      const newMovieError = { ...prev };
+
+      newMovieError[field] = error;
+
+      return newMovieError;
+    });
+  };
+
+  const handleOnChange = (field: keyof Movie, newValue: string): void => {
```
Viktor-Kost, line 54: and add `type` to `e`
```suggestion
  const handleOnChange = (e): void => {
```

Viktor-Kost (approved):
Nice!

PR #2263, opened 2024-03-27
Solution1

Salmaralex:
[DEMO LINK](https://Salmaralex.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,86 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import React from 'react';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [formData, setFormData] = useState({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+
+  const handleTextFieldChange = (name: string) => (value: string) => {
+    setFormData(prevData => ({
+      ...prevData,
+      [name]: value,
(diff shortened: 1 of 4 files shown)


l4st1m0za (changes requested):
Good job but you need to make some fixes:
 
 1) Button for adding the movie is always active, even if I have not all fields field with text
 2) After submitting the movie, all the fields start to look like this
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,99 @@
...
 
       <TextField
         name="title"
-        label="Title"
-        value=""
-        onChange={() => {}}
-        required
+        label="Name"
+        value={formData.title}
+        required={true}
```
l4st1m0za, line 50: There is no need to add {true} to the required attribute, simply writing required is enough, as it implies true.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,99 @@
...
+  const handleTextFieldChange = (name: string) => (value: string) => {
+    setFormData(prevData => ({
+      ...prevData,
+      [name]: value,
+    }));
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (!formData.imgUrl || !formData.imdbUrl || !formData.imdbId) {
```
l4st1m0za, line 28: Here the form is only checking the imgUrl, imdbUrl, and imdbId fields for being empty, while the title field is also required and should be checked.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,99 @@
...
+    <form className="NewMovie" onSubmit={handleSubmit}>
+      <h2 className="title">Добавить фильм</h2>
 
       <TextField
         name="title"
-        label="Title"
-        value=""
-        onChange={() => {}}
-        required
+        label="Name"
```
l4st1m0za, line 48: The label for the title field is 'Name', it should be 'Title' for better consistency.

l4st1m0za (approved):
Good job!

PR #2267, opened 2024-03-31
task solution

BoGoBooM:
[DEMO LINK](https://bogoboom.github.io/react_movies-list-add-form/)


anastasiiavorobiova (changes requested):
Great work! Consider adding a movie to the end of the list 
Screenshot 2024-03-31 at 16 53 08
Consider trimming values to avoid adding a movie with spaces only

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,102 @@
...
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys: don't hesitate to ask for some help with that in the chat

etojeDenys (changes requested):
prev comment still not fixed :(
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,111 @@
...
+
+    if (isValid) {
+      onAdd({
+        title: newMovie.title.trim(),
+        description: newMovie.description.trim(),
+        imgUrl: newMovie.imgUrl.trim(),
+        imdbUrl: newMovie.imdbUrl.trim(),
+        imdbId: newMovie.imdbId.trim(),
+      });
+    }
```
andriy-shymkiv, line 47: I still can add a movie with only spaces:
image
you need to trim the values, and then check if they are `true`, and only then - add the movie, otherwise - block the submit button

andriy-shymkiv (changes requested):

maxim2310 (approved):
Well done!!!

PR #2268, opened 2024-03-31
React Movies List Add Form

Denys209818:


Denys209818:
[DEMO LINK](https://denys209818.github.io/react_movies-list-add-form)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,179 @@
...
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
   const [count] = useState(0);
+  const [movie, setMovie] = useState<MovieCustom>({
+    title: { value: '', touched: false },
+    description: { value: '', touched: false },
+    imdbId: { value: '', touched: false },
+    imdbUrl: { value: '', touched: false },
+    imgUrl: { value: '', touched: false },
+  });
```
etojeDenys, line 39: you can move touched logic to the TextField and create there touched state, after that, just save the string here in the fields of this object(`title: ''`)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,179 @@
...
+        onChange={value => {
+          setMovie(prev => {
+            return {
+              ...prev,
+              imgUrl: { touched: prev.imgUrl.touched, value: value },
+            };
+          });
+        }}
+        changeTouched={changeTouched}
+        required={true}
```
etojeDenys, line 135:
```suggestion
        required
```
we can simplify it

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,179 @@
...
+        });
+
+        setMovie({
+          title: { value: '', touched: false },
+          description: { value: '', touched: false },
+          imdbId: { value: '', touched: false },
+          imdbUrl: { value: '', touched: false },
+          imgUrl: { value: '', touched: false },
+        });
+      }}
```
etojeDenys, line 87: create some function for that

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,179 @@
...
-        required
+        value={movie.title}
+        onChange={value => {
+          setMovie(prev => {
+            return {
+              ...prev,
+              title: { touched: prev.title.touched, value: value },
+            };
+          });
+        }}
```
etojeDenys, line 102: same here move this logic to some function

etojeDenys (changes requested):
image

src/components/NewMovie/NewMovie.tsx
Denys209818: Ok, then provide an idea how to reset all of textfields after submiting a form

etojeDenys: image
I don't understand what you mean, you reset the text boxes here or what?

etojeDenys (changes requested):
don't hesitate to tag me in the chat and ask for some help with that

src/components/NewMovie/NewMovie.tsx
Denys209818: Yes, I need do it. Because touched inside a TextView will be true after submiting the form. Therefore, value is empty touched is true. There is an error. I need to hide it. And to do it, I need set all touched states as false. I can't do it from parental component. Then I thought that I can control it from parental component.

volodymyr-soltys97 (approved):
Nice work 🔥

PR #2270, opened 2024-04-01
solution

KachVl:
[DEMO LINK](https://kachvl.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -6,22 +6,24 @@ type Props = {
   onAdd: (movie: Movie) => void;
 };
 
+const emptyInfo = {
+  title: '',
+  description: '',
+  imgUrl: '',
+  imdbUrl: '',
+  imdbId: '',
+};
+
 export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
   const [count, setCount] = useState(0);
-  const [newMovie, setNewMovie] = useState({
-    title: '',
-    description: '',
-    imgUrl: '',
-    imdbUrl: '',
-    imdbId: '',
-  });
+  const [newMovie, setNewMovie] = useState({ ...emptyInfo });
 
-  const newMovieInputs = (value: string, key: string) => {
+  const addNewMovieData = (value: string, name: string) => {
     setNewMovie({
       ...newMovie,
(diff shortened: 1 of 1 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,141 @@
...
+      imdbUrl: value,
+    });
+  };
+
+  const newMovieImdbId = (value: string) => {
+    setNewMovie({
+      ...newMovie,
+      imdbId: value,
+    });
+  };
```
etojeDenys, line 54: try to create only one function that will change the field value for all fields

etojeDenys (changes requested):

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,113 @@
...
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [newMovie, setNewMovie] = useState({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
```
Viktor-Kost, line 19: move an object to a variable outside the component so you can reuse this object at `reset` function

src/components/TextField/TextField.tsx
```diff
@@ -45,7 +45,7 @@ export const TextField: React.FC<Props> = ({
           })}
           placeholder={placeholder}
           value={value}
-          onChange={event => onChange(event.target.value)}
+          onChange={event => onChange(event.target.value, name)}
```
Viktor-Kost, line 48:
```suggestion
          onChange={onChange}
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,113 @@
...
+
+  const newMovieInputs = (value: string, key: string) => {
+    setNewMovie({
+      ...newMovie,
+      [key]: value,
+    });
+  };
```
Viktor-Kost, line 26: 1. function name should start with a verb
2. and add a type to `e`
```suggestion
  const addNewMovie = (e: ...) => {
    const { name, value } = e.target;
    
    setNewMovie({
      ...newMovie,
      [name]: value,
    });
  };
```

Viktor-Kost (approved):
Nice!

PR #2275, opened 2024-04-04
solution

Maksym-Atanasov:
[DEMO LINK](https://Maksym-Atanasov.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,86 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  addMovie: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ addMovie }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imageUrl, setImageUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  function resetForm() {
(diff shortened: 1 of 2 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The fields cannot be an empty. You need to fix it
image
image

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,99 @@
...
+export const NewMovie: React.FC<Props> = ({ addMovie }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imageUrl, setImageUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys: don't hesitate to ask for some help in the chat

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,99 @@
...
           <button
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={
+              !title.trim() ||
+              !imageUrl.trim() ||
+              !imdbId.trim() ||
+              !imdbUrl.trim()
+            }
```
etojeDenys, line 96: better to move it to some var

etojeDenys (changes requested):

src/components/NewMovie/NewMovie.tsx
Maksym-Atanasov: it's only make code harder to understand

TarasHoliuk: Makes sense in this case since your TextField component handles event, and `onChange` prop is a function that accepts actually input text.
But the approach Denys suggested sometimes could be handy - it allows to create only 1 function for different inputs

PR #2278, opened 2024-04-05
cope

true-binar-White-Rabbit:
[DEMO LINK](https://true-binar-White-Rabbit.github.io/react_movies-list-add-form/)


src/components/TextField/TextField.tsx
```diff
@@ -20,14 +20,33 @@ export const TextField: React.FC<Props> = ({
...
+      '|(?:www\\.|[-;:&=+$,\\w]+@)[A-Za-z0-9.-]+)((?:\\/[+~%/.\\w-_]*)?' +
+      '?\\??(?:[-+=&;%@,.\\w_]*)#?(?:[,.!/\\\\\\w]*))?)$',
+  );
+
+  const hendleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setFieldValue(e.target.value);
+    onChange(e.target.value, name);
+
+    if (name === 'imgUrl' || name === 'imdbUrl') {
+      if (!pattern.test(e.target.value) && required) {
```
VitaliyBondarenko1982, line 43: validate value after onBlur event. it should fix these two tests
Screenshot 2024-04-05 at 13 22 34

src/components/TextField/TextField.tsx
```diff
@@ -20,14 +20,33 @@ export const TextField: React.FC<Props> = ({
...
-  // generage a unique id once on component load
   const [id] = useState(() => `${name}-${getRandomDigits()}`);
-
-  // To show errors only if the field was touched (onBlur)
+  const [fieldValue, setFieldValue] = useState(value);
   const [touched, setTouched] = useState(false);
-  const hasError = touched && required && !value;
+  const [hasUrlError, setHasUrlError] = useState(false);
+
+  const hasError = (touched && required && !fieldValue) || hasUrlError;
```
VitaliyBondarenko1982, line 30:
```suggestion
  const hasError = (touched && required && !fieldValue.trim()) || hasUrlError;
```
now no any error if type only spaces in input. trim fieldValue for show error if in input all characters are spaces.

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comments and try to fix tests and linter warnings.

src/components/TextField/TextField.tsx
true-binar-White-Rabbit: I added additional validation to check the correct url. For this reason, your test is not successful, but if the 'clever' test is a priority for you, then I will leave the default option. It is much easier to just refer to the test than to delve into the approach to solving the problem that the student has applied and help optimize the code that already performs the task. thank you

anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2024-04-05 at 14 59 51
Consider showing errors and disabling the submit if there are just spaces here

true-binar-White-Rabbit:
pls check code , this is already the case

anastasiiavorobiova (approved):
Awesome work!

PR #2281, opened 2024-04-06
solution

pasha1932:
[the demo page](https://pasha1932.github.io/react_movies-list-add-form/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The fields cannot be a empty, you need to fix it
image
image

SanyaBratashchuk (approved):

PR #2283, opened 2024-04-07
add task solution

Romanov7788:
[DEMO LINK](https://romanov7788.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,15 +2,23 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { useState } from 'react';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  const addMovie = (movie: Movie) => {
+    setMovies(movies.concat(movie));
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addMovie} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,42 +1,111 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
-
-  return (
-    <form className="NewMovie" key={count}>
-      <h2 className="title">Add a movie</h2>
-
-      <TextField
-        name="title"
-        label="Title"
-        value=""
-        onChange={() => {}}
-        required
-      />
-
-      <TextField name="description" label="Description" value="" />
-
-      <TextField name="imgUrl" label="Image URL" value="" />
-
-      <TextField name="imdbUrl" label="Imdb URL" value="" />
-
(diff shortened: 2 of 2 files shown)


src/App.tsx
```diff
@@ -2,15 +2,23 @@ import './App.scss';
...
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { useState } from 'react';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  const addMovie = (movie: Movie) => {
+    setMovies(movies.concat(movie));
```
Viktor-Kost, line 12: 1. use `prev` values
2. you can use spread instead of  `concat`
```suggestion
    setMovies(prevMovies => ({ ...prevMovies, movie }));
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,42 +1,111 @@
...
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState<string>('');
+  // const [hasTitleError, setHasTitleError] = useState<boolean>(false);
```
Viktor-Kost, line 12: remove comments

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,42 +1,111 @@
...
+  const [imdbUrl, setImdbUrl] = useState<string>('');
+  // const [hasImdbUrlError, setHasImdbUrlError] = useState<boolean>(false);
+
+  const [imdbId, setImdbId] = useState<string>('');
+  // const [hasImdbIdError, setHasImdbIdError] = useState<boolean>(false);
 
-  return (
-    <form className="NewMovie" key={count}>
-      <h2 className="title">Add a movie</h2>
+  const checkFields = () => {
```
Viktor-Kost, line 25: 1. add `trim` method to avoid passing just `spaces`
2. it's not a function, you can just use as a variable
Screenshot 2024-04-07 at 21 07 40

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,42 +1,111 @@
...
+
+  const [description, setDescription] = useState<string>('');
+
+  const [imgUrl, setImgUrl] = useState<string>('');
+  // const [hasImgUrlError, setHasImgUrlError] = useState<boolean>(false);
+
+  const [imdbUrl, setImdbUrl] = useState<string>('');
+  // const [hasImdbUrlError, setHasImdbUrlError] = useState<boolean>(false);
+
+  const [imdbId, setImdbId] = useState<string>('');
```
Viktor-Kost, line 22: It's not advisable to generate such a large number of states when we only have one instance (movie). 

```
const initialMovieState = {
  title: '',
  description: '',
  imgUrl: '',
  imdbUrl: '',
  imdbId: ''
}
```

4. Whenever you need to change value of any field you should pass only `e` and then just change a value of a movie object by `name` key. In this case you will reuse one handler instead of handlers for each input
Example: 

```
  setMovie(prevInputs => ({ ...prevInputs, [name]: value }))
}
```

```suggestion
  const [movie, setMovie] = useState(initialMovieState);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,42 +1,111 @@
...
+      imdbId: imdbId,
+    };
 
-      <TextField name="imgUrl" label="Image URL" value="" />
+    onAdd(movie);
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
```
Viktor-Kost, line 45: create a `reset` function where you just pass an `initialState`

Viktor-Kost (changes requested):

etojeDenys (changes requested):
image

etojeDenys (approved):
lgtm

PR #2290, opened 2024-04-10
solution

weswojciech:
https://weswojciech.github.io/react_movies-list-add-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,15 +2,23 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { useState } from 'react';
+import { Movie } from './types/Movie';
 
 export const App = () => {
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -2,15 +2,23 @@ import './App.scss';
...
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { useState } from 'react';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = (movie: Movie) => {
+    setMovies(movies.concat(movie));
```
nataliaklonowska, line 12: when updating the state based on the previous state’s value always use function that receives the previous state as its argument
```suggestion
    setMovies(prevMovies => prevMovies.concat(movie));
```

nataliaklonowska (changes requested):

nataliaklonowska (approved):

PR #2293, opened 2024-04-11
Solution proposed

bartKarb:
[DEMO LINK](https://bartKarb.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,15 +2,22 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { useState } from 'react';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+  const addMovie = (movie: Movie) => {
+    setMovies(movies.concat(movie));
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -2,15 +2,22 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { useState } from 'react';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+  const addMovie = (movie: Movie) => {
+    setMovies(movies.concat(movie));
```
nataliaklonowska, line 11: when updating the state based on the previous state’s value always use function that receives the previous state as its argument

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,84 @@
...
+
+  const handleSubmit: FormEventHandler = e => {
+    e.preventDefault();
+    const movie = {
+      title: title,
+      description: description,
+      imgUrl: imgUrl,
+      imdbUrl: imdbUrl,
+      imdbId: imdbId,
+    };
```
nataliaklonowska, line 27: if your object keys and values are the same you can use object property shorthand :)
https://javascript.plainenglish.io/object-literals-using-object-property-shorthand-6360825c60ef

nataliaklonowska (changes requested):

nataliaklonowska (approved):

PR #2296, opened 2024-04-12
Solution

kubsen:
[DEMO LINK](https://kubsen.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,15 +2,23 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { useState } from 'react';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = (movie: Movie) => {
+    setMovies(movies.concat(movie));
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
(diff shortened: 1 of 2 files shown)


choeqq (changes requested):
URL field should accept only URL data `http://something.x`

src/App.tsx
```diff
@@ -2,15 +2,23 @@ import './App.scss';
...
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { useState } from 'react';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = (movie: Movie) => {
+    setMovies(movies.concat(movie));
```
nataliaklonowska, line 12: when updating the state based on the previous state’s value always use function that receives the previous state as its argument

```suggestion
    setMovies(prevMovies => prevMovies.concat(movie));
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,92 @@
...
+
+  const handleSubmit: FormEventHandler = e => {
+    e.preventDefault();
+    const movie = {
+      title: title,
+      description: description,
+      imgUrl: imgUrl,
+      imdbUrl: imdbUrl,
+      imdbId: imdbId,
+    };
```
nataliaklonowska, line 29: if your object keys and values are the same you can use object property shorthand :)
https://javascript.plainenglish.io/object-literals-using-object-property-shorthand-6360825c60ef

nataliaklonowska (changes requested):

nataliaklonowska (approved):

PR #2311, opened 2024-04-16
task solution

Kacper7001:
[DEMO LINK](https://Kacper7001.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -54,13 +54,7 @@ export const NewMovie: React.FC<NewMovieAdd> = ({ onAdd }) => {
     const isValidImgUrl = isValidUrl(imgUrl);
     const isValidImdbUrl = isValidUrl(imdbUrl);
 
-    if (
-      !title ||
-      !description ||
-      !isValidImgUrl ||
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,139 @@
...
+    setDescription(event.target.value);
+  };
+
+  const handleSetImdbUrl = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setImdbUrl(event.target.value);
+  };
+
+  const handleSetImgUrl = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setImgUrl(event.target.value);
+  };
```
choeqq, line 33: please make sure that urls are actually urls, u have to validate them

choeqq (changes requested):

choeqq (changes requested):
@Kacper7001 did you test this solution? I can't type inside url inputs 🤔

nataliaklonowska (changes requested):
description field shouldn't be required (that's why your tests fail)

nataliaklonowska (approved):

PR #2323, opened 2024-04-23
add task solution

valost:
https://valost.github.io/react_movies-list-add-form/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
 import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+import { Movie } from './types/Movie';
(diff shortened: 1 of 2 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The add button doesn't work
image
image

I can't add a new movie
image

valost:
But I can add a new movie with the valid links
Screenshot 2024-04-24 at 11 07 47

valost:
So the add button is disabled unless I insert a valid link to the input field

volodymyr-soltys97 (changes requested):
If the input is not valid, then you do not need to disable the button, but you need to show the text next to these inputs that they are not valid. It is necessary that the user understands why there is no possibility to add a movie.
For example like this
image

etojeDenys (approved):
lgtm

PR #2327, opened 2024-04-27
add_solution

metrakonon:
[DEMO](https://metrakonon.github.io/react_movies-list-add-form/)


ArtemZakhar (approved):
I like your solution!)
But I would not surprised If mentor asked you to make one input state in shape of an object)))
Great job)

PR #2332, opened 2024-04-30
add task solution

JohnyJDX:
[DEMO LINK](https://JohnyJDX.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,103 @@
 import { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+interface Props {
+  onAdd: (movie: Movie) => void;
+}
+
+const pattern =
+  // eslint-disable-next-line max-len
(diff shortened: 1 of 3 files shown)


yanamomot (approved):
Good, approved!

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,111 @@
...
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+interface Props {
+  onAdd: (movie: Movie) => void;
+}
+
+const pattern =
+  // eslint-disable-next-line max-len
```
undeletable, line 10: [Not critical, recommendation for future work]
Instead of disabling `max-len` rule for this line, you can change its configuration to ignore regular expressions: https://eslint.org/docs/latest/rules/max-len#ignoreregexpliterals

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,111 @@
 import { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+interface Props {
```
undeletable, line 5: It's better to give this interface a more clear name, e.g. `NewMovieProps`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,111 @@
...
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+    setCount(count + 1);
+  };
 
   return (
-    <form className="NewMovie" key={count}>
+    <form className="NewMovie" key={count} onSubmit={handleSubmit}>
```
undeletable, line 60: This is a single rendered element, so `key` prop is not required here

undeletable (approved):
The code looks fine! I just have several advices oh how to improve it - you can keep them in mind during the further work.

PR #2336, opened 2024-05-01
task solution

kinqbert:
[DEMO LINK](https://kinqbert.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,25 @@
 import './App.scss';
+
+import { useState } from 'react';
+
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
 
 export const App = () => {
+  const [visibleMovies, setVisibleMovies] = useState(moviesFromServer);
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={visibleMovies} />
       </div>
       <div className="sidebar">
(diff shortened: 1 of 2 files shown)


PrytulaAlexandr (approved):
Hello Maksym. Everything is fine, but i have a little coment for you. Check please number 3 of task list Disable the submit button until all the required fields are filled (spaces should be trimmed); Your demo works some little diferent).

kinqbert:
> Hello Maksym. Everything is fine, but i have a little coment for you. Check please number 3 of task list Disable the submit button until all the required fields are filled (spaces should be trimmed); Your demo works some little diferent).

Thanks for noticing! Don't know how did I miss that. Fixed my code :)

kinqbert:
Thank you for pointing out errors! Implemented fixes into my code 🤝

src/components/TextField/TextField.tsx
```diff
@@ -21,13 +22,23 @@ export const TextField: React.FC<Props> = ({
...
+
   const [touched, setTouched] = useState(false);
-  const hasError = touched && required && !value;
+  const hasError =
+    (touched && required && !value.trim()) ||
+    (touched && value && !matchesPattern);
+  let errorMessage = `${label} is required`;
+
+  if (!matchesPattern && value) {
+    errorMessage = 'Valid URL expected';
```
alexander-ignatow, line 40: I would live TextField universal component, so probably this message should also be set up via props, or you should make it more generic - like `The value is not valid` (preferably do both)

alexander-ignatow (approved):
Good job!

I like how you updated the TextField - it is almost universal, just see my comment on the error message for the future

PR #2340, opened 2024-05-02
Solution

odunchyk:
[DEMO LINK](https://odunchyk.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,87 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+interface Props {
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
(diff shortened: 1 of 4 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -27,7 +27,7 @@ export const TextField: React.FC<Props> = ({
 
   // To show errors only if the field was touched (onBlur)
   const [touched, setTouched] = useState(false);
-  const hasError = touched && required && !value;
+  const hasError = touched && required && !value.trim();
```
alexander-ignatow, line 30: The text field appears universal component, so it is a bad idea to put this here

If you want to trim something - do it outside the TextField component

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,95 @@
...
+        onChange={newTitle => setTitle(newTitle)}
+        required={true}
       />
 
-      <TextField name="description" label="Description" value="" />
+      <TextField
+        name="description"
+        label="Description"
+        value={description}
+        onChange={newDescription => setDescription(newDescription)}
```
alexander-ignatow, line 59: You may try just 
```suggestion
        onChange={setDescription}
```

alexander-ignatow (changes requested):
Good, but please, fix some minor issues

alexander-ignatow (approved):
Good job!

👍

PR #2342, opened 2024-05-02
add solution

D-Kondratiev:
[the demo page](https://D-Kondratiev.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,88 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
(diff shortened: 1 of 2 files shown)


OTsiolkovskiy (approved):
Good job!

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,96 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
+import { Movie } from '../../types/Movie';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
```
mshepel, line 5: It's better to use more expressive name for this type, so that it's clear what exactly it means. For example, `NewMovieProps`.

mshepel (changes requested):
Everything looks good, except a single issue. 

Steps to reproduce;
1. Enter some value for a required field
2. Switch to a different field
3. Switch back to field from step 1
4. Remove value from this field

Result:
Error message is displayed immediately

Expected result:
Error message should only be displayed when field loses focus

You might have to modify `TextField`  component to fix this. Also in `hasError` condition in `TextField` component it makes sense to check if trimmed value is valid.

src/components/NewMovie/NewMovie.tsx
undeletable, line 5: It's better to use more expressive name for this type, so that it's clear what exactly it means. For example, `NewMovieProps`.

undeletable (changes requested):
Everything looks good, except a single issue. 

Steps to reproduce;
1. Enter some value for a required field
2. Switch to a different field
3. Switch back to field from step 1
4. Remove value from this field

Result:
Error message is displayed immediately

Expected result:
Error message should only be displayed when field loses focus

You might have to modify `TextField`  component to fix this. Also in `hasError` condition in `TextField` component it makes sense to check if trimmed value is valid.

PR #2354, opened 2024-05-03
add solving

ordila:
[DEMO LINK](https://ordila.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,66 @@
...
-        label="Title"
-        value=""
-        onChange={() => {}}
-        required
-      />
+  const fields = getFields(form);
 
-      <TextField name="description" label="Description" value="" />
+  const updateForm = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { name, value } = event.target;
```
ArtemZakhar, line 22: cheeter)))

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,66 @@
...
+    setForm(INITIAL_FORM);
+
+    setCount(prevCount => prevCount + 1);
+  };
+
+  return (
+    <form className="NewMovie" key={count} onSubmit={handleSubmitForm}>
+      <h2 className="title">Add a movie</h2>
+      {fields.map(field => (
+        <TextField key={field.name} {...field} onChange={updateForm} />
```
ArtemZakhar, line 55: It is awesome!)

src/components/TextField/TextField.tsx
```diff
@@ -1,33 +1,32 @@
...
   required = false,
   onChange = () => {},
 }) => {
-  // generage a unique id once on component load
   const [id] = useState(() => `${name}-${getRandomDigits()}`);
 
-  // To show errors only if the field was touched (onBlur)
   const [touched, setTouched] = useState(false);
-  const hasError = touched && required && !value;
+  const hasError = touched && required && !value.trim();
```
ArtemZakhar, line 30: ))))

ArtemZakhar (approved):
It was like a good book!)
Very sad that the story ends(

src/helpers/getTextField.tsx
```diff
@@ -0,0 +1,22 @@
...
+      name: 'description',
+      label: 'Description',
+      value: description,
+      required: false,
+    },
+    { name: 'imgUrl', label: 'Image URL', value: imgUrl, required: true },
+    { name: 'imdbUrl', label: 'Imdb URL', value: imdbUrl, required: true },
+    { name: 'imdbId', label: 'Imdb ID', value: imdbId, required: true },
+  ];
+};
```
alexander-ignatow, line 22: It is a good idea to break all into parts, but sometimes it is ok to use simpler file structure

For instance - this can be safely placed inside NewMoviwForm, as you don't use (and don't plan to) anywhere outside the NewMovieForm

But it is up to you fix it or not

src/constants/form.constants.ts
```diff
@@ -0,0 +1,7 @@
+export const INITIAL_FORM = {
+  title: '',
+  description: '',
+  imgUrl: '',
+  imdbId: '',
+  imdbUrl: '',
+};
```
alexander-ignatow, line 7: This is clearly part of the NewMovieForm logic - what is the reason to put it in some different folder

src/components/TextField/TextField.tsx
```diff
@@ -1,33 +1,32 @@
...
```
alexander-ignatow, line 29: You should not do so, as `TextField` is a universal input, and potentially it can be used in different forms with different logic

If you need to trim, do this outside the `TextField` itself, or at least make this behavior optional (and not default)

alexander-ignatow (changes requested):
Very interesting approach

Please see my comments, as some of the pointed issues need to be fixed

Actually, you handle the fields just like it is done in many real-world apps 👍 

Still, there are a few issues.

And most important - don't try to think too far ahead, because your solution is too universal

For the future - always think "Am I really going to re-use this code" or "How exactly will I need this code in the future?"

Actually, in practice, it is often considered to be "ok" (but not good) to have one copy-paste :)

But when you are copy-pasting code for the second time (i.e. have three copies of the same code in total) you should refactor the solution.

src/helpers/getTextField.tsx
ordila: I just try to use harder way for more practice , thank's a lot

mrkovaliv (approved):
👍

PR #2359, opened 2024-05-06
Solution

lmmaarriiaa:
[DEMO LINK](https://lmmaarriiaa.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,113 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isUrlValid = (url: string) => {
+    // eslint-disable-next-line
+    const pattern = /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
+
+    if (!url) {
+      return true;
+    }
+
+    return pattern.test(url);
+  };
+
+  const isFormValid =
+    !!title.trim() &&
+    !!imgUrl.trim() &&
+    !!imdbUrl.trim() &&
+    !!imdbId.trim() &&
+    isUrlValid(imdbUrl) &&
+    isUrlValid(imgUrl);
+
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,122 @@
...
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isUrlValid = (url: string) => {
+    const pattern =
+      // eslint-disable-next-line
```
alexander-ignatow, line 21: if you have to disable eslint, it is better to disable the single rule (or few rules) then the whole check

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,122 @@
...
+      /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
+
+    if (!url) {
+      return true;
```
alexander-ignatow, line 25: Maybe `false` :)

As an empty string is definitely not a valid url

src/components/TextField/TextField.tsx
```diff
@@ -8,6 +8,7 @@ type Props = {
   placeholder?: string;
   required?: boolean;
   onChange?: (newValue: string) => void;
+  isUrlValid?: boolean;
```
alexander-ignatow, line 11: This is good, you try adding this, but in this very case, it makes sense to make `TextField` more generic, and more universal

1/3

```suggestion
  externalError?: string
```

src/components/TextField/TextField.tsx
```diff
@@ -41,7 +43,7 @@ export const TextField: React.FC<Props> = ({
           id={id}
           data-cy={`movie-${name}`}
           className={classNames('input', {
-            'is-danger': hasError,
+            'is-danger': hasError || !isUrlValid,
```
alexander-ignatow, line 46: 2/3

```suggestion
            'is-danger': hasError || externalError,
```

src/components/TextField/TextField.tsx
```diff
@@ -51,6 +53,10 @@ export const TextField: React.FC<Props> = ({
       </div>
 
       {hasError && <p className="help is-danger">{`${label} is required`}</p>}
+
+      {!hasError && !isUrlValid && (
+        <p className="help is-danger">Please, enter correct url</p>
+      )}
```
alexander-ignatow, line 59: 3/3
```suggestion

      {!hasError && externalError && (
        <p className="help is-danger">{externalError}</p>
      )}
```

alexander-ignatow (approved):
Very good!

Lets go with it, but for the future please see my comments for the TextField

src/components/NewMovie/NewMovie.tsx
lmmaarriiaa: In my TextField component I added a additional red text `'Please, enter correct url'` which os shown when there is url and it is `!isUrlValid` and I dont want this text to be shown when there is no url. Thats why true here :)

src/components/TextField/TextField.tsx
lmmaarriiaa: Got it, thanks

src/components/TextField/TextField.tsx
lmmaarriiaa: Got it!

lmmaarriiaa: Thank you for your review, very useful comments!

PR #2360, opened 2024-05-06
add task solution

Follder:
[DEMO LINK](https://follder.github.io/react_movies-list-add-form/)

Everything works, but I can`t pass one test 'should add a movie with a correct data'.
And I don`t know how I can fix it (

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,88 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isFormValid =
+    !!title && !!imgUrl && !!imdbUrl && !!imdbId && imdbUrl.includes(imdbId);
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const newMovie: Movie = {
+      title,
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    };
+
+    onAdd(newMovie);
+
+    setCount(current => current + 1);
+
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+  };
 
   return (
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,96 @@
...
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
+  const isFormValid =
+    !!title && !!imgUrl && !!imdbUrl && !!imdbId && imdbUrl.includes(imdbId);
```
lmmaarriiaa, line 20: `imdbUrl.includes(imdbId)`
Smart, but kind of unnecessary check here, at first I even thought that your demo isnt working :)

lmmaarriiaa (approved):
Hi, Dmytro! 
Good job!
I am not sure, but maybe your problem is connected with the fact that you are adding a new movie not in the end of a list, but in the start. 
Please, try to change your approach

src/components/TextField/TextField.tsx
```diff
@@ -45,7 +45,7 @@ export const TextField: React.FC<Props> = ({
           })}
           placeholder={placeholder}
           value={value}
-          onChange={event => onChange(event.target.value)}
+          onChange={event => onChange(event.target.value.trimStart())}
```
alexander-ignatow, line 48: This is not good

TextField is a universal component, and you should not just change it's functionality

If you need to trim this, do it outside the TextField

src/components/NewMovie/NewMovie.tsx
alexander-ignatow, line 20: This check makes sense, but it is a bad user experience - you just block the `Add` button without any messages

You need to either remove this check, or to add some clear message why the button is blocked

alexander-ignatow (changes requested):
Good, but some issues are needed to be fixed

Follder:
Show an error if one of the field are empty
Show an error if Imdb URL dont include Imdb ID
Make trim outside the TextField component

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,42 +1,128 @@
...
+    !!imgUrl.trim() &&
+    !!imdbUrl.trim() &&
+    !!imdbId.trim() &&
+    !!imdbUrl.trim();
+
+  const isValidImdbId = imdbUrl.includes(imdbId);
+
+  const isFormValid = isFieldsNotEmpty && isValidImdbId;
+
+  useEffect(() => {
```
mrkovaliv, line 31: move this logic to custom hook

mrkovaliv:
It isn't good practice show error on first render, when user doesn't make any actions.
Show error after click on button

Screenshot 2024-05-27 at 18 08 12

mrkovaliv (changes requested):
Good job, but there are some improvements)))

Follder:
Solved the problem. Now the error is not shown on the first render, only after clicking the button.

But now the tests are wrong, because in doc, the button should be disabled by default.

But I agree with [alexander-ignatow](https://github.com/alexander-ignatow), that it can be bad practice.

mrkovaliv (approved):
👍 👍 👍

PR #2361, opened 2024-05-06
Develop

Olek-91:
[DEMO LINK](https://Olek-91.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,37 +1,143 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
+import { InputState } from '../../types/InputState';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (movie: Movie) => void;
+}
+
+type State = {
+  titleValue: string;
+  imgUrlValue: string;
+  imdbUrlValue: string;
+  imdbIdValue: string;
+  descrValue: string;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const initialValue = {
+    titleValue: '',
+    imgUrlValue: '',
+    imdbUrlValue: '',
+    imdbIdValue: '',
+    descrValue: '',
+  };
+
+  const [count, setCount] = useState(0);
+  const [inputValue, setInputValue] = useState<State>(initialValue);
+
+  const { titleValue, imgUrlValue, imdbUrlValue, imdbIdValue, descrValue } =
+    inputValue;
+
+  const pattern =
+    // eslint-disable-next-line
+    /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
+
+  const notAllInformationGathered = Object.entries(inputValue).some(
+    ([key, value]) => {
+      if (key === 'descrValue') {
+        return false;
+      }
+
+      return value.length === 0;
+    },
+  );
+
+  const resetInputState = () => {
+    setInputValue(() => {
+      return {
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,143 @@
...
+  const [count, setCount] = useState(0);
+  const [inputValue, setInputValue] = useState<State>(initialValue);
+
+  const { titleValue, imgUrlValue, imdbUrlValue, imdbIdValue, descrValue } =
+    inputValue;
+
+  const pattern =
+    // eslint-disable-next-line
+    /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
+
```
ilyasolo881, line 36: Nice! Use pattern, i forgot about it :)

ilyasolo881 (approved):
Nice work, look's good!

src/components/TextField/TextField.tsx
```diff
@@ -1,13 +1,15 @@
...
 
 type Props = {
   name: string;
   value: string;
   label?: string;
   placeholder?: string;
   required?: boolean;
-  onChange?: (newValue: string) => void;
+  onChange?: (newValue: string, key: InputState) => void;
+  inputState: InputState;
```
alexander-ignatow, line 12: This is a bad idea, as you make TextInput dependent on the properties of some form

Please think another way to do this (maybe use `name`) ?

src/types/InputState.ts
```diff
@@ -0,0 +1,7 @@
+export enum InputState {
+  titleValue = 'titleValue',
+  imgUrlValue = 'imgUrlValue',
+  imdbUrlValue = 'imdbUrlValue',
+  imdbIdValue = 'imdbIdValue',
+  descrValue = 'descrValue',
+}
```
alexander-ignatow, line 7: You do not need this
See: https://www.typescriptlang.org/docs/handbook/2/keyof-types.html

alexander-ignatow (changes requested):
Good job, however some fixes are needed

image

It is good you check the URLs, but the user experience is not perfect - the button looks working, but when the URL is not valid, it just silently does nothing.

Some other issues are needed to be fixed too

src/components/TextField/TextField.tsx
```diff
@@ -45,7 +43,7 @@ export const TextField: React.FC<Props> = ({
           })}
           placeholder={placeholder}
           value={value}
-          onChange={event => onChange(event.target.value)}
+          onChange={event => onChange(event.target.value.trimStart())}
```
undeletable, line 46: It's not a good practice to change the input value itself.
When I type something into text field, and see that it contains something different from what I typed, it feels like app is buggy.
I'd suggest not to do anything with field value itself, but instead change the way how you calculate `hasError`.

src/components/TextField/TextField.tsx
```diff
@@ -22,10 +22,8 @@ export const TextField: React.FC<Props> = ({
   required = false,
   onChange = () => {},
 }) => {
-  // generage a unique id once on component load
   const [id] = useState(() => `${name}-${getRandomDigits()}`);
 
-  // To show errors only if the field was touched (onBlur)
   const [touched, setTouched] = useState(false);
   const hasError = touched && required && !value;
```
undeletable, line 28: According to the condition, we should display error message when the field loses focus.
Let's consider the following set of steps:
1. Type something into text field
2. Switch to another field
3. Switch back to the field from step 1

Error message would be displayed immediately, while it should be there only when you focus on a different field. That's because `touched` is always true after first blur event. Please fix that.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,114 @@
...
+      <TextField
+        name="imgUrl"
+        label="Image URL"
+        value={imgUrl}
+        onChange={setImgUrl}
+        required
+      />
 
-      <TextField name="imdbUrl" label="Imdb URL" value="" />
+      <TextField
```
undeletable, line 89: As Oleksandr suggested, when the URL field value is an invalid URL, it makes sense to add error message under the input about that. Probably you'd need to enhance `TextField` behavior to handle that.
Same for image URL field.

undeletable (changes requested):
Looks good overall, but in some cases the behavior of your app could be improved. Please check the comments.

PR #2362, opened 2024-05-07
Solution

ilyasolo881:
[DEMO LINK](https://ilyasolo881.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,88 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,96 @@
...
+      description,
+      imgUrl,
+      imdbUrl,
+      imdbId,
+    });
+
+    resetAll();
+  };
+
+  const validValue =
```
tarassamchenko, line 41: better to use question-name like 'isFormValid'

tarassamchenko (approved):
Good job! Keep it up

src/components/NewMovie/NewMovie.tsx
alexander-ignatow: makes sense - mind the naming

alexander-ignatow (approved):
Good, but mind the naming

PR #2363, opened 2024-05-07
movies-list-add-form

ksstassy:
[DEMO LINK](https://ksstassy.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,93 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,97 @@
...
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
```
l4st1m0za, line 16: Instead of all those states, you could also user single state with object inside, which would safe all tha data needed

l4st1m0za (approved):
Good job!

PR #2368, opened 2024-05-08
add task solution

ArtemDemeniev:
https://artemdemeniev.github.io/react_movies-list-add-form/


stepanyukvova200 (approved):
Good job!
Everything works, and the code looks pretty fine

PR #2370, opened 2024-05-08
solution

opalahecha:
[DEMO LINK](https://opalahecha.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,127 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type State = {
+  title: string;
+  imgUrl: string;
+  imdbUrl: string;
+  imdbId: string;
+  description: string;
+};
+
+export const NewMovie: React.FC<{ onAdd: (movie: Movie) => void }> = ({
+  onAdd,
+}) => {
+  const [count, setCount] = useState(0);
+  const [inputValue, setInputValue] = useState<State>({
+    title: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+    description: '',
+  });
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [errors, setErrors] = useState<Partial<State>>({});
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,140 @@
...
+  imdbUrl: string;
+  imdbId: string;
+  description: string;
+};
+
+export const NewMovie: React.FC<{ onAdd: (movie: Movie) => void }> = ({
+  onAdd,
+}) => {
+  const [count, setCount] = useState(0);
+  const [inputValue, setInputValue] = useState<State>({
```
etojeDenys, line 17:
```suggestion
  const [inputValue, setInputValue] = useState<Movie>({
```
looks like we can simplify it

etojeDenys (changes requested):
image
I can create the movie only with spaces, let's fix that

IvanFesenko (changes requested):
Why I can't add a movie? If there any validation and input is not valid - please show error message
https://github.com/mate-academy/react_movies-list-add-form/assets/61904995/146300e4-e86a-42e8-b721-0fbb7e94b948

opalahecha:
@IvanFesenko 
Probably because of this part of the code. When I enter imgUrl and imdbUrl from https://www.imdb.com/, it works. Maybe I should have added this check in a different way. If I remove this part, it works with everything.  Delete this part?
const pattern =
      // eslint-disable-next-line max-len
      /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
    if (!fields()) {
      return;
    }
    const { imgUrl, imdbUrl } = inputValue;
    if (!imgUrl.match(pattern) || !imdbUrl.match(pattern)) {
      return;
    }
   
a

etojeDenys (changes requested):
better to just show the error message near to some fields that are invalid
image
or don't hesitate to ask for some help in the chat

anastasiiavorobiova (approved):
Great work!

PR #2373, opened 2024-05-12
add task solution

DenLysiak:
[DEMO LINK](https://DenLysiak.github.io/react_movies-list-add-form/)


vadiimvooo (changes requested):
Good job. There is one bug. You can add movie with empty title and Imdb ID. Please fix it.
image

maxim2310 (approved):
Well done!

PR #2380, opened 2024-05-19
add task solution

wasik07:
[DEMO LINK](https://wasik07.github.io/react_movies-list-add-form/)


src/services/Pattern.ts
```diff
@@ -0,0 +1,7 @@
+export function getPatternCorect(strValue: string): boolean {
```
SanyaBratashchuk, line 1:
```suggestion
export function isValidUrl(strValue: string): boolean {
```

PR #2382, opened 2024-05-21
Solution

przwojwwp:
[DEMO LINK](https://przwojwwp.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,15 +2,23 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import React, { useState } from 'react';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = React.useCallback((newMovie: Movie) => {
+    setMovies(prevMovies => [...prevMovies, newMovie]);
+  }, []);
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addMovie} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,105 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import React from 'react';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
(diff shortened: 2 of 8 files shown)


src/App.tsx
```diff
@@ -2,15 +2,28 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import React, { useState } from 'react';
+
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = (newMovie: {
```
yevhenii-pyl, line 10: Wouldn't it be easier to create a type for `Movie` and use it everywhere instead of manually typing each property it has? What if Movie as a type will be extended in future? :)

src/App.tsx
```diff
@@ -2,15 +2,28 @@ import './App.scss';
...
+    title: string;
```
yevhenii-pyl, line 11: Btw, it looks like a decent place to use useCallback in order to create a referential equality

For more info: https://kentcdodds.com/blog/usememo-and-usecallback

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,124 @@
...
+  onAdd: (newMovie: FormValues) => void;
+};
+
+export const NewMovie: React.FC<NewMovieProps> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const [formValues, setFormValues] = useState<FormValues>({
```
yevhenii-pyl, line 22: A cleaner approach would be to create a `defaultFormValues` or something.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,124 @@
...
+  });
+
+  const handleChange = (fieldName: keyof FormValues, value: string) => {
+    setFormValues(prevState => ({
+      ...prevState,
+      [fieldName]: value,
+    }));
+  };
+
+  const isFormValid = ['title', 'imgUrl', 'imdbUrl', 'imdbId'].every(
```
yevhenii-pyl, line 37: Solid approach, but why don't we create a 'requiredFields' array? :)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,124 @@
...
+    event.preventDefault();
+    if (!isFormValid) {
+      return;
+    }
+
+    // Add the new movie
+    onAdd(formValues);
+
+    // Clear the form
+    setFormValues({
```
yevhenii-pyl, line 51: If only you had a `defaultFormValues` to use here :)

yevhenii-pyl (changes requested):
Good work!
Let's just adjust a few minor things to make your code even cleaner and we are good to go!

yevhenii-pyl (approved):
You got it! 🥇

PR #2384, opened 2024-05-21
add task solution

sophiasph:
[DEMO LINK](https://sophiasph.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -22,10 +22,10 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   };
 
   const visibleButton = !(
-    newMovie.title &&
-    newMovie.imgUrl &&
-    newMovie.imdbUrl &&
-    newMovie.imdbId
+    newMovie.title.trim() &&
+    newMovie.imgUrl.trim() &&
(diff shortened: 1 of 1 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,101 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImgbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys: feel free to ask for some help with that in the chat

etojeDenys (changes requested):

vadiimvooo (changes requested):
Good job, but you have one thing to fix. For now it's possible to add movie with empty fields which contains only spaces. Please fix it.
image

PR #2386, opened 2024-05-22
solution

YuliiaCat:
https://yuliiacat.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,91 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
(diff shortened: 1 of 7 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,99 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys: feel free to ask for some help in the chat

etojeDenys (changes requested):

vadiimvooo (changes requested):
Good job, but you have to show errorsfor 3 components on blur to pass the tests. Please feel free to ask for help in fe_chat.
`   1) should show imgUrl error only after blur
    2) should show imdbUrl error only after blur
    3) should show imdbId error only after blur`

PR #2387, opened 2024-05-23
task solved

meljaszuk:
[DEMO LINK](https://meljaszuk.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -57,7 +57,7 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
         label="Description"
         value={description}
         onChange={setDescription}
-        required
+        required={false}
       />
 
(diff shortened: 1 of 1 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,96 @@
...
         required
       />
 
-      <TextField name="description" label="Description" value="" />
+      <TextField
+        name="description"
+        label="Description"
+        value={description}
+        onChange={setDescription}
+        required
```
yevhenii-pyl, line 60: This field should be optional: 

`The description is optional;`

yevhenii-pyl (changes requested):
Good job! 
Could you adjust that one tiny thing with description field just to make it right?

BogdanMaliuta (approved):
Nice!

PR #2394, opened 2024-05-28
react_movies-list-add-form

Donovanoff:
[DEMO LINK](https://Donovanoff.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
 import { Movie } from '../../App';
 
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,97 @@
...
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
anastasiia-tilikina, line 17: let's combine these states into one

first let's create a var with initial values

```typescript
const initialValues = {
  title: ''
  // ...
``` 
then create one state

```typescript
const [movieInfo, setMovieInfo] = useState(initialValues)
``` 
and then we can create only one handler for all fields

```javascript
const handleInputChange = (e: //... ) => {
  const {name, value } = e.target;
  
  setMovieInfo(prev => ({
    ...prev,
  [name]: value,
  }))
```

anastasiia-tilikina (changes requested):
Well done!

let's make improvement here

etojeDenys (approved):
lgtm

PR #2405, opened 2024-06-08
react list movies

HryniukTaras:
[DEMO LINK](https://HryniukTaras.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,100 @@
...
+type Props = {
+  onAdd: (newMovie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
VitaliyBondarenko1982, line 14:
```suggestion
  const [newMovie, setNewMovie] = useState({
     title: '',
     description: '',
     ...
  })
```
suggest to make one state for all form fields. use event.target.name for change proper field in newMovie state (each input should have name attribute with same value as newMovie appropriate key).

VitaliyBondarenko1982 (changes requested):
Good job. Please check comment

and also pay attention on such case:
now i can fill input with all spaces and no see any error, but submit button still disabled.

Screenshot 2024-06-08 at 20 33 55

anastasiiavorobiova (approved):
Great work!

PR #2409, opened 2024-06-12
add App newMovie

VvynnykV:
[DEMO LINK](https://VvynnykV.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,101 @@
-import { useState } from 'react';
-import { TextField } from '../TextField';
+import React, { useState } from 'react';
 
-export const NewMovie = () => {
+import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
+
+type Props = {
+  onAdd: (newMovie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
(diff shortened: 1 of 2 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,109 @@
...
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
vadiimvooo, line 19: you can merge this useStates into one
```suggestion
  const [newMovie, setNewMovie] = useState({
     title: '',
     description: '',
     ...
  })
```

vadiimvooo: For changing proper field in this useState you can use handle function
```
  const handleChange = (name: keyof Movie, value: string) => {
    setNewMovie(prev => ({
      ...prev,
      [name]: value,
    }));
  };
```

vadiimvooo (changes requested):
Good job, you have one thing to fix.

src/components/NewMovie/NewMovie.tsx
VvynnykV: I have done changes, but in my opinion its became more complicate to understand) especially with function handleChange

vadiimvooo (approved):
Well done

PR #2434, opened 2024-06-26
Add movies form

GoVolodya:
[DEMO LINK](https://GoVolodya.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,95 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [disabled, setDisabled] = useState(true);
+
+  if (!disabled && (!title || !imgUrl || !imdbUrl || !imdbId)) {
+    setDisabled(true);
+  }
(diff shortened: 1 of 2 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The fields cannot be only spaces, you need to fix it 
image
image

GoVolodya:
Have made a fix for this issue. Now spaces are not counted as word.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,109 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):

src/components/NewMovie/NewMovie.tsx
GoVolodya: Nice suggestion, it's better to have less useStates

src/types/Movie.ts
```diff
@@ -1,4 +1,5 @@
 export interface Movie {
+  [index: string]: string;
```
etojeDenys, line 2: why do you need it, looks redundant

etojeDenys (changes requested):

src/types/Movie.ts
GoVolodya: No, it is not. For some reasons Typescript gives an error if you do something like this
`for (const key in newMovie) {
      newMovie[key] = newMovie[key].trim();
  }`
  while deploying the project.
  
  To fix this errors I had to add this line. Found in [Typescript documentation](https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures)

etojeDenys: it's a bad practice, better try to ask for some help with that error in the chat

etojeDenys (changes requested):

src/types/Movie.ts
GoVolodya: okay, I have removed that.

GoVolodya:
Have made some improvements for this task. Now code looks better.

SanyaBratashchuk (approved):

PR #2435, opened 2024-06-26
add task solution

kravets111:
[DEMO LINK](https://kravets111.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,143 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,156 @@
...
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState('');
```
volodymyr-soltys97, line 12: Use only one state for the movie 
```suggestion
  const [movie, setMovie] = useState({
    title: '',
    ...
  });
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better
1. Use trim method for the form fields to remove the invalidity text if we have spaces at the beginning or end
image

etojeDenys (approved):
lgtm

PR #2448, opened 2024-07-03
add solution

om1gse:


src/App.tsx
```diff
@@ -2,15 +2,23 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
```
seriuksergii, line 2: You are importing a file 'MoviesList' but it is not in the folder 'components'

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,34 +1,77 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
```
seriuksergii, line 2: You are importing a file 'TextField' but it is not in the 'src'

seriuksergii (changes requested):
You should add DEMO LINK in description your pull request.

PR #2451, opened 2024-07-06
Solution

kyrylomanko:

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,7 +10,7 @@ export const App = () => {
         <MoviesList movies={moviesFromServer} />
       </div>
       <div className="sidebar">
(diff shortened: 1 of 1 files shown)


VitaliyBondarenko1982 (changes requested):
- Can't to type in input. Looks like name attribute wasn't settled to input.
- disable add button when not valid at least one of inputs fields.

VitaliyBondarenko1982 (changes requested):
Almost done.
Please check such case:
when fill input only with empty spaces - no visible any errors, but add button still disabled.
movies_form

kyrylomanko:
Made it so you cannot use space as first symbol

PR #2460, opened 2024-07-17
add NewMovie FC

OlehPhw:
[DEMO LINK](https://OlehPhw.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,94 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,102 @@
...
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [valueTitle, setValueTitle] = useState('');
+  const [valueDesc, setValueDesc] = useState('');
+  const [valueImg, setValueImg] = useState('');
+  const [valueUrl, setValueUrl] = useState('');
+  const [valueId, setValueId] = useState('');
```
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):
try to fix your demo link

volodymyr-soltys97 (approved):
Nice work 🔥

PR #2464, opened 2024-07-28
solution

manch0ffline:
[DEMO LINK](https://manch0ffline.github.io/react_movies-list-add-form/)


src/components/TextField/TextField.tsx
```diff
@@ -44,7 +44,7 @@ export const TextField: React.FC<Props> = ({
             'is-danger': hasError,
           })}
           placeholder={placeholder}
-          value={value}
+          value={value.trim()}
```
VitaliyBondarenko1982, line 47: try to type more than one word in title or description. it is impossible now .
the reason is here
```suggestion
          value={value}
```

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comment and try to pass all tests.

vadiimvooo (approved):
Well done.

PR #2477, opened 2024-08-06
Solution

Teollan:
[DEMO LINK](https://Teollan.github.io/react_movies-list-add-form/)

- [x] Url checker implemented


Teollan:
I have no idea wtf is wrong with automated tests, but it works on my machine...

image
image

MaxBorovique (approved):
Great!

src/App.tsx
```diff
@@ -2,15 +2,24 @@ import './App.scss';
...
     <div className="page">
       <div className="page-content">
         <MoviesList movies={moviesFromServer} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie
+          onAdd={movie => {
+            moviesFromServer.push(movie);
+            requestRerender();
```
ArturErmolenko, line 20: There is one rule that you should always remember
DO NOT mutate data
I believe if you set movies from server in state and than work with state you absolutely don't need this kostyls))
When you change state --> component rerender

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,127 @@
...
+    };
+
+    onAdd(movie);
+
+    // reset
+    setTitle({ value: '', isValid: false });
+    setDescription({ value: '', isValid: false });
+    setImgUrl({ value: '', isValid: false });
+    setImdbUrl({ value: '', isValid: false });
+    setImdbId({ value: '', isValid: false });
```
ArturErmolenko, line 65: can we make reset function?

src/components/TextField/TextField.tsx
```diff
@@ -21,13 +22,30 @@ export const TextField: React.FC<Props> = ({
...
+    }
+
+    return inputValidator(toCheck);
+  };
+
+  const errorMessage = checkForErrorMessage(value);
+  const isValid = errorMessage === '';
+
+  // on mount determine if initial value is valid
+  useEffect(() => onChange(value, isValid), []);
```
ArturErmolenko, line 45: dependencies?

src/utils/regexp.ts
```diff
@@ -0,0 +1,3 @@
+export const urlPattern =
+  // eslint-disable-next-line max-len
```
ArturErmolenko, line 2: use + for this string, disable linter it's a bad practice

ArturErmolenko (changes requested):
GJ 👍 
Let's improve a bit)

src/utils/regexp.ts
Teollan: Let`s just say, that linter rules dont apply in a file that only has one regexp in it... 

Because I have no idea how to split this pattern **and** make it work.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,129 @@
...
 
-      <TextField name="imgUrl" label="Image URL" value="" />
+      <TextField
+        onChange={(value, isValid) => setImgUrl({ value, isValid })}
+        name="imgUrl"
+        label="Image URL"
+        value={imgUrl.value}
+        inputValidator={value =>
+          isUrl(value) ? '' : `${value} is not a valid url`
+        }
```
DanilWeda, line 97: This can be moved to func

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,129 @@
...
+        required
+      />
 
-      <TextField name="imdbUrl" label="Imdb URL" value="" />
+      <TextField
+        onChange={(value, isValid) => setImdbUrl({ value, isValid })}
+        name="imdbUrl"
+        label="Imdb URL"
+        value={imdbUrl.value}
+        inputValidator={value =>
```
DanilWeda, line 106: Then use in this place ( to prevent code duplication)

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,129 @@
...
+  onAdd?: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd = () => {} }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const [title, setTitle] = useState<StringWithValidation>({
```
DanilWeda, line 18: For the future: you can unit some states into one state by using an object in useState.

DanilWeda (approved):
Good job. 
Approved with small comments. 
Many thanks)

PR #2524, opened 2024-09-06
done

v-zagorovskii:
[DEMO LINK](https://v-zagorovskii.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -52,7 +52,7 @@ export const NewMovie: React.FC<Props> = ({ onAdd }) => {
       <TextField
         newName="title"
         label="Title"
-        newValue={newMovie.title.trim()}
+        newValue={newMovie.title.trimStart()}
         onChange={(name: string, value: string) =>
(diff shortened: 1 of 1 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,94 @@
...
-  const [count] = useState(0);
+type Props = {
+  onAdd: (newMovie: Movie) => void;
+};
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [title, setTitle] = useState('');
+  const [description, setDescrip] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
etojeDenys, line 13: we can combine it in one useState
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):
I can create the movie just only with spaces, let's fix that
image

vadiimvooo (approved):
LGTM

PR #2539, opened 2024-09-21
add task solution

modeltoIT:
[DEMO LINK](https://modeltoIT.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,10 +1,22 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+interface Props {
+  onAdd: (movies: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -13,25 +25,66 @@ export const NewMovie = () => {
...
+      />
 
       <div className="field is-grouped">
         <div className="control">
           <button
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!allFilled}
+            onClick={event => {
```
VitaliyBondarenko1982, line 71: make funciton in component and pass reference here

src/components/NewMovie/NewMovie.tsx
```diff
@@ -13,25 +25,66 @@ export const NewMovie = () => {
...
+                imgUrl: imgURL,
+                imdbId: imdbID,
+                imdbUrl: imdbURL,
+              });
+              setTitle('');
+              setImgURL('');
+              setImdbURL('');
+              setDescription('');
+              setImdbID('');
+              setCount(count + 1);
```
VitaliyBondarenko1982, line 86: suggest to make resetForm function and just call here

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some improvements.

Take a look on such behavior

https://github.com/user-attachments/assets/ef9098e6-a862-471b-b738-cb12baffeb0f

As user a can't see any error messages when fill input with empty spaces.

vadiimvooo (approved):
Lgtm

PR #2551, opened 2024-10-02
solution

YuliiaShatkovska:
[DEMO LINK](https://YuliiaShatkovska.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,88 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+type Props = {
+  onAdd: (newMovie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [count, setCount] = useState(0);
+  const [titleValue, setTitleValue] = useState('');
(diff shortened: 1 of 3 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,96 @@
...
+    event.preventDefault();
+
+    if (isValidFields) {
+      onAdd({
+        title: titleValue,
+        description: descriptionValue || '',
+        imgUrl: imgUrlValue,
+        imdbUrl: imdbUrlValue,
+        imdbId: imdbIdValue,
+      });
```
hma-3, line 38: 1. You must trim values before adding (I saw that you trim at the beginning, but don't forget the end)
2. there is no point in writing `descriptionValue || ''` -- if the user has not written a description 'descriptionValue' will have default value, which is already equal to `''`

hma-3 (approved):
LGTM
Have just a few suggestions below

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,96 @@
...
+        onChange={value => setTitleValue(value)}
         required
       />
 
-      <TextField name="description" label="Description" value="" />
+      <TextField
+        name="description"
+        label="Description"
+        value={descriptionValue}
+        onChange={value => setDescriptionValue(value)}
```
VolodymyrKirichenko, line 60: it will be the same. For all cases
```suggestion
        onChange={setDescriptionValue}
```

src/components/NewMovie/NewMovie.tsx
VolodymyrKirichenko: it's true.
Do it

VolodymyrKirichenko (approved):
Good job, but plz, review my comments :)

PR #2562, opened 2024-10-04
Some tests are not passed

Pidsvynok:
image
https://<your_account>.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,99 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+interface Props {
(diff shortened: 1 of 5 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,107 @@
...
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+
+  const { description, ...valuesWithoutDescription } = formValues;
+  const formHasEmptyFields = Object.values(valuesWithoutDescription).some(
+    value => value === '',
```
etojeDenys, line 23: I can create the movie just with spaces, let's fix that here
image

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #2563, opened 2024-10-04
Done

MaxF1996:
- [DEMO LINK](https://MaxF1996.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,103 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+export const NewMovie = ({
+  onAdd = () => {},
+}: {
+  onAdd?: (movie: Movie) => void;
+}) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+
+  const [currentTitle, setCurrentTitle] = useState('');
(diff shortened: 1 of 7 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,110 @@
...
+        imdbId: currentImdbId.trim(),
+      });
+
+      setCurrentTitle('');
+      setCurrentDescription('');
+      setCurrentImdbUrl('');
+      setCurrentImdbId('');
+      setCurrentImgUrl('');
+
+      setCount(count + 1);
```
VitaliyBondarenko1982, line 54: make resetForm function to reset inputs and update count. use prev state for update count state

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,110 @@
...
+        onChange={title => setCurrentTitle(title)}
         required
       />
 
-      <TextField name="description" label="Description" value="" />
+      <TextField
+        name="description"
+        label="Description"
+        value={currentDescription}
+        onChange={description => setCurrentDescription(description)}
```
VitaliyBondarenko1982, line 74:
```suggestion
        onChange={setCurrentDescription}
```
it can be shorter. same for other inputs

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comments and also pay attention on such case:
- I can fill required inputs with empty spaces but no see any error (title, imdbId).
Expected behavior - should show error on blur event.


https://github.com/user-attachments/assets/9f647470-c1e8-4e86-bf4d-b4caa78b641d

vadiimvooo (approved):
Well done

PR #2572, opened 2024-10-07
Add task solution

WatarJoy:
https://watarjoy.github.io/react_movies-list-add-form/

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,101 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
-  // Increase the count after successful form submission
-  // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+interface Props {
+  onAdd: (movie: Movie) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [title, setTitle] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [description, setDescription] = useState('');
+  const [errors, setErrors] = useState({
+    title: false,
+    imgUrl: false,
+    imdbUrl: false,
+    imdbId: false,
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,109 @@
...
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [description, setDescription] = useState('');
+  const [errors, setErrors] = useState({
+    title: false,
+    imgUrl: false,
+    imdbUrl: false,
+    imdbId: false,
+  });
```
YuriiMykhailenko, line 20: You don't need that much hooks just use one which stores an object. And you don't need to create state for errors with object in it. Actually you have already 'required' in the input elements and it works so your implementation duplicates code and error message, just check if required field are not empty and according to that make it disabled or not. 
```suggestion
export const NewMovie: React.FC<Props> = ({ onAdd }) => {
  const [newFilm, setNewFilm] = useState<FilmType | null>(null);
```

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,109 @@
...
+      />
+      {errors.imdbId && <p className="help is-danger">IMDB ID is required</p>}
 
       <div className="field is-grouped">
         <div className="control">
           <button
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!title || !imgUrl || !imdbUrl || !imdbId}
```
YuriiMykhailenko, line 106: make a variable for that
```suggestion
            disabled={isReadyToSubmit}
```

src/components/TextField/TextField.tsx
```diff
@@ -46,7 +46,10 @@ export const TextField: React.FC<Props> = ({
           placeholder={placeholder}
           value={value}
           onChange={event => onChange(event.target.value)}
-          onBlur={() => setTouched(true)}
+          onBlur={() => {
+            setTouched(true);
+            onBlur();
```
YuriiMykhailenko, line 51: remove this onBlur()  it's the cause of duplicated message

YuriiMykhailenko (changes requested):
You did fine job👍 But I have to request some changes, because mentors will request them too, pay attention to my comments below

PR #2603, opened 2024-11-07
solution 1.0

StanislavKapytsia:
# Movies list - Add Form

[DEMO LINK](https://StanislavKapytsia.github.io/react_movies-list-add-form/)

You have the `App` with the `MoviesList` and `NewMovie` form containing ready
to use `TextField` components. Learn how it works and implement an ability to
add movies from [IMDB](https://www.imdb.com/).

If you want to test your page you can get first image from a [movie page](https://www.imdb.com/title/tt1312171) using `DevTools` -> `Network` -> `Img`

> Here is [the demo page](https://mate-academy.github.io/react_movies-list-add-form/)

1. `NewMovie` should check if `title`, `imgUrl`, `imdbUrl`, `imdbId` are
entered when an input looses focus (`onBlur`) and show an error and a red
border if needed (learn how it it implemented in the `TextField`);
1. The `description` is optional;
1. Disable the submit button until all the required fields are filled (spaces should be trimmed);
1. Clear the form after adding a new movie.
1. Errors should not be shown after clearing the form (change its key to
reinitialize the form);

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,115 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
 
-export const NewMovie = () => {
+type Move = {
+  title: string;
+  description: string;
+  imgUrl: string;
+  imdbUrl: string;
+  imdbId: string;
+};
+
+interface Props {
+  onAdd: (movie: Move) => void;
+}
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,123 @@
...
+  }
+
+  function getImdb(data: string) {
+    setImdb(data);
+  }
+
+  function getId(data: string) {
+    setId(data);
+  }
+
```
etojeDenys, line 46: we can combine it in one useState
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):
1. I can create the movie just with spaces, let's fix that
image

2.

etojeDenys (approved):
🔥

PR #2621, opened 2024-11-26
solution

minimal005:
[DEMO LINK](https://minimal005.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,24 @@
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+import { Movie } from './types/Movie';
 import moviesFromServer from './api/movies.json';
+import { useState } from 'react';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const changedListOfMovies = (movie: Movie) => {
+    setMovies(currentMovies => [...currentMovies, movie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,16 +1,24 @@
...
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
+import { Movie } from './types/Movie';
 import moviesFromServer from './api/movies.json';
+import { useState } from 'react';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const changedListOfMovies = (movie: Movie) => {
```
StasSokolov1, line 11: will be better use name like `addNewMovie`

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,107 @@
...
+enum Fields {
+  title = 'Title',
+  description = 'Description',
+  imgUrl = 'Image URL',
+  imdbUrl = 'Imdb URL',
+  imdbId = 'Imdb ID',
+}
+
+export const NewMovie: React.FC<Props> = props => {
+  const { onAdd } = props;
```
StasSokolov1, line 22:
```suggestion
export const NewMovie: React.FC<Props> = { onAdd } => {
```

can be a bit simplify

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,107 @@
...
           <button
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={
+              !isValid(title) ||
+              !isValid(imgUrl) ||
+              !isValid(imdbUrl) ||
+              !isValid(imdbId)
+            }
```
StasSokolov1, line 104: let's move this into constant because looks a bit complicated

StasSokolov1 (approved):
Almost done, let's fix a few comments before approving

PR #2638, opened 2024-12-04
solution

Fridrif27:
[DEMO LINK](https://fridrif27.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,15 +2,23 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { NewMovie } from './components/NewMovie';
 import moviesFromServer from './api/movies.json';
+import { useState } from 'react';
+import { Movie } from './types/Movie';
 
 export const App = () => {
+  const [movies, setMovies] = useState(moviesFromServer);
+
+  const addMovie = (movie: Movie) => {
+    setMovies(prevMovies => [...prevMovies, movie]);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={addMovie} />
       </div>
     </div>
   );
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,96 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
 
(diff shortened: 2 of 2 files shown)


src/App.tsx
```diff
@@ -2,15 +2,21 @@ import './App.scss';
...
+
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
       </div>
       <div className="sidebar">
-        <NewMovie /* onAdd={(movie) => {}} */ />
+        <NewMovie onAdd={(movie: Movie) => addMovie(movie)} />
```
AnderZerfall, line 19: You can use {addMovie} instead of (movie: Movie) => addMovie(movie).  React will automatically pass the value

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,93 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+interface Props {
```
AnderZerfall, line 5: Not a big deal, though you could use type instead of interface

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,93 @@
...
-    <form className="NewMovie" key={count}>
+    <form className="NewMovie" key={count} onSubmit={handleSubmit}>
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
```
AnderZerfall, line 48: The same thing here: you don't have to pass value explicitly. React will do that automatically for you

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,93 @@
...
+        required
+      />
 
       <div className="field is-grouped">
         <div className="control">
           <button
             type="submit"
             data-cy="submit-button"
             className="button is-link"
+            disabled={!title || !imgUrl || !imdbUrl || !imdbId}
```
AnderZerfall, line 90: I'm uncertain about this issue for a couple of reasons. First, it seems like a waste of space. Second, you didn't take into account that a single space (' ') will evaluate to TRUE just like any other text. I suggest creating a separate validation function to eliminate unnecessary spaces (you can use the trim() function for this) and then check if the resulting strings are not empty.

AnderZerfall (changes requested):
Hey, great job! I left some comments below to help you improve your code quality.

PR #2665, opened 2024-12-08
Solution

IlMarauder:
https://IlMarauder.github.io/react_movies-list-add-form/


OkMoroz (changes requested):
Unfortunately, the link _https://ilmarauder.github.io/react_movies-list-add-form/_ is not working (404 error), so I wasn't able to view your page. I also noticed that not all tests passed. I’m forced to reject it. Sorry.

PR #2708, opened 2025-01-13
solution

djtyrf312:
[DEMO LINK](https://djtyrf312.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,30 +1,95 @@
-import { useState } from 'react';
+import React, { useState } from 'react';
 import { TextField } from '../TextField';
+import { Movie } from '../../types/Movie';
 
-export const NewMovie = () => {
+type Props = {
+  onAdd: (newValue: Movie) => void;
+};
+
+export const NewMovie = ({ onAdd }: Props) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [description, setDescription] = useState('');
+  let isDisabled = true;
+
+  if (title.trim() && imgUrl.trim() && imdbUrl.trim() && imdbId.trim()) {
+    isDisabled = false;
+  }
+
+  const movie = {
+    title,
+    description,
+    imgUrl,
+    imdbUrl,
+    imdbId,
+  };
+
+  const clearForm = () => {
+    setTitle('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+    setDescription('');
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    onAdd(movie);
+    clearForm();
+    setCount(count + 1);
+  };
 
   return (
-    <form className="NewMovie" key={count}>
+    <form
+      className="NewMovie"
+      key={count}
+      onSubmit={handleSubmit}
+    >
       <h2 className="title">Add a movie</h2>
 
       <TextField
         name="title"
         label="Title"
-        value=""
-        onChange={() => {}}
+        value={title}
+        onChange={value => setTitle(value)}
         required
       />
 
-      <TextField name="description" label="Description" value="" />
+      <TextField
+        name="description"
+        label="Description"
+        value={description}
+        onChange={value => setDescription(value)}
+      />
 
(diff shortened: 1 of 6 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,113 @@
...
+
+function isValidUrl(url: string) {
+  const pattern =
+    // eslint-disable-next-line max-len
+    /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
+
+  return pattern.test(url);
+}
+
+export const NewMovie = ({ onAdd }: Props) => {
```
naziamoff: I'd call it `NewMovieForm` so it is clear we're dealing with a form

src/components/TextField/TextField.tsx
```diff
@@ -8,26 +8,31 @@ type Props = {
...
   onChange?: (newValue: string) => void;
+  isValidUrl?: (url: string) => boolean;
 };
 
 function getRandomDigits() {
   return Math.random().toFixed(16).slice(2);
 }
 
+
+
```
naziamoff: usually we need no more than one empty line

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,113 @@
...
+      />
 
-      <TextField name="imdbUrl" label="Imdb URL" value="" />
+      <TextField
+        name="imdbUrl"
+        label="Imdb URL"
+        value={imdbUrl}
+        required
+        onChange={value => setImdbUrl(value)}
+        isValidUrl={value => isValidUrl(value)}
```
naziamoff: by the name `isValidUrl` I think that we're dealing with `boolean`, but we're passing a function. To make it more obvious we're dealing with a function, its name should start with a verb `get/check/find/validate` etc.

src/components/TextField/TextField.tsx
```diff
@@ -50,7 +55,13 @@ export const TextField: React.FC<Props> = ({
         />
       </div>
 
-      {hasError && <p className="help is-danger">{`${label} is required`}</p>}
+      {hasEmptyFieldError && (
+        <p className="help is-danger">{`${label} is required`}</p>
+      )}
+
+      {hasWrongUrlError && (
+        <p className="help is-danger">{`Url is incorrect`}</p>
```
naziamoff: It seems you don't need to wrap it into jsx expression, it is just a text

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,113 @@
...
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [description, setDescription] = useState('');
+  let isDisabled = true;
+
+  if (title.trim() && imgUrl.trim() && imdbUrl.trim() && imdbId.trim()) {
+    isDisabled = false;
+  }
+
+  const movie = {
```
naziamoff: it seems that this variable is only used for passing in `onAdd`, but I might be missing sth out. If that's the case - you can pass it as an object right away.

```
onAdd({
  title,
  description,
  imgUrl
  ...otherStuff
})
```
The way you did it is not considered a mistake, but I'm just a huge fan of a shorter syntax where possible. Up to you

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,113 @@
...
+  const [title, setTitle] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+  const [description, setDescription] = useState('');
+  let isDisabled = true;
+
+  if (title.trim() && imgUrl.trim() && imdbUrl.trim() && imdbId.trim()) {
+    isDisabled = false;
+  }
```
naziamoff: Looks like a constant value, since we only reassign it once. Here's my suggestion. Looks cleaner to me this way, since all of our values have the same validation logic, so we can avoid repeating `x.trim()`

```suggestion
  const isDisabled = [title, imgUrl, imdbUrl, imdbId, description]
    .some((value) => !value.trim()):
```

naziamoff (approved):

src/components/NewMovie/NewMovie.tsx
djtyrf312: nice

src/components/NewMovie/NewMovie.tsx
djtyrf312: changed to validateUrl

PR #2726, opened 2025-01-27
implement new movie

yzhyhaliuk:
[DEMO LINK](https://yzhyhaliuk.github.io/react_movies-list-add-form/)
I'm not sure about the pattern, it's not working when I test it with proper URL


anastasiiavorobiova (approved):
Great work!

PR #2802, opened 2025-03-17
Solution

DariaRadtkina:
https://DariaRadtkina.github.io/react_movies-list-add-form/


AnOvramenko (approved):
Good job! 👍 
The inputs should not work with only spaces. ©AiBuddy)

PR #2883, opened 2025-05-06
add task solution

Uusuff:
-[DEMO LINK](https://Uusuff.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -8,11 +8,14 @@ type Props = {
 
 export const NewMovie: React.FC<Props> = ({ onAdd }) => {
   const [formKey, setFormKey] = useState(0);
-  const [title, setTitle] = useState('');
-  const [description, setDescription] = useState('');
-  const [imgUrl, setImgUrl] = useState('');
-  const [imdbUrl, setImdbUrl] = useState('');
-  const [imdbId, setImdbId] = useState('');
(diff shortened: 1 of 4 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,118 @@
...
+  onAdd: (movie: Movie) => void;
+};
+
+export const NewMovie: React.FC<Props> = ({ onAdd }) => {
+  const [formKey, setFormKey] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
```
etojeDenys: we can combine it in one useState
```
  title: '',
  ...
})
```

```
}
```

etojeDenys (changes requested):

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The url is valid now, you need to use `trim()` method and hide the error 
image

Uusuff:
Снимок экрана 2025-05-09 173038

When I run the test, everything works correctly, the photo is attached. I added code to remove spaces via the trim() method
 
 urlPattern.test(value.trim()) ? null : 'Must be a valid URL'

anastasiiavorobiova (approved):
Awesome!

PR #2936, opened 2025-05-29
add task solution

Khmara-Danylo:
[DEMO LINK](https://Khmara-Danylo.github.io/react_movies-list-add-form/)

- [x] The `NewMovie` component should verify that the `title`, `imgUrl`, `imdbUrl`, and `imdbId` fields are filled out when an input field loses focus (`onBlur`). If any of these fields are empty, it should display an error message and apply a red border to indicate the issue. This functionality is already implemented in the `TextField`, so you can refer to that for guidance on how it works. There is no need to implement this part in this task;
- [x] The `description` is optional;
- [x] Disable the submit button until all the required fields are filled (spaces should be trimmed);
- [x] Clear the form after adding a new movie.
- [x] Errors should not be shown after clearing the form (change its key to
reinitialize the form);

- [x] (Optional) Advanced validation


RTSaitama (commented):
gg wp

RTSaitama (approved):

PR #2944, opened 2025-05-30
Update MovieCard.tsx

Dikanb:


RTSaitama (changes requested):
some fail with PR happened, as ive seen.  Waitin for right changes

PR #3031, opened 2025-08-08
sol

O1ympus:
[DEMO LINK](https://O1ympus.github.io/react_movies-list-add-form/)


MykytaMusaiev (approved):
Good job!
For next time you can use single state for one form, like:
const [form, setForm] = useState({
title: '',
description '',
imgUrl: '',
imdbUrl: '',
imdbId: ''
});

And work with that! Would be a bit easier, with less useStates in youd code in my personal opinion.

Keep it up mate!

PR #3043, opened 2025-08-14
solution for react movies list add form task

Aonarro:
https://aonarro.github.io/react_movies-list-add-form/


PopelinovDanilo (approved):
good job

PR #3061, opened 2025-08-17
Solution

iastivka:
[DEMO LINK](https://iastivka.github.io/react_movies-list-add-form/)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,37 +1,89 @@
...
+    setForm(prev => ({ ...prev, [field]: value }));
+  };
+
+  const isValid =
+    form.title.trim() &&
+    form.imgUrl.trim() &&
+    form.imdbUrl.trim() &&
+    form.imdbId.trim();
 
+  const handleSubmit = (event: React.FormEvent) => {
```
horidor, line 31: 5. Errors should not be shown after clearing the form (change its key to reinitialize the form);

Task template has count variable and count key in the form tag - you should use it to reset error fields.

horidor (changes requested):
Good job! You need to complete part 5 of the task, as your project doesn't clear form errors on submit.

horidor (approved):
Good job!

PR #3135, opened 2025-10-15
solution

voloshynmykyta:
[DEMO LINK](https://voloshynmykyta.github.io/react_movies-list-add-form/)


MaksimShe:
u can add film with not valid info 
btw, in ReadMe we don`t have any info about - what do with no valid info. i think u do this task very well!

https://github.com/user-attachments/assets/9be0a715-36ad-49ca-9071-44e801270707

MaksimShe (approved):

PR #3147, opened 2025-10-17
add task solution

JackSavin:
[DEMO LINK](https://jacksavin.github.io/react_movies-list-add-form/)


S-Khraban (approved):
Use the Classnames library.
It’s not explicitly mentioned in the task, but mentors require it.

PR #3209, opened 2026-01-01
Completed task

Augustwise:
The `value` attribute was changed to `defaultValue`.

The following functions are used for form validation:

* `normalize` (to remove spaces)
* `validateTitle` - the title is considered valid if it contains from 2 to 40 characters
* `validateDescription` - the description is considered valid if it contains from 25 to 400 characters (if the user entered it)
* `isHttpUrl` - this function validates a URL. It uses the `URL` interface, which greatly simplifies validation. Regular expressions are not used
* `validateRequiredUrl` - in this function the URL is validated in the same way, using the `isHttpUrl` function.

The `createEmptyMovieForm` function returns a ready “movie” object (entity)

The validity of all values is checked as follows:

```js
const errors = {
  title: validateTitle(form.title),
  description: validateDescription(form.description),
  imgUrl: validateRequiredUrl(form.imgUrl, 'Image URL'),
  imdbUrl: validateRequiredUrl(form.imdbUrl, 'Imdb URL'),
};

const isFormValid = Object.values(errors).every(message => !message);
```

In `<p className="help is-danger">`, a custom error is now used: `{errorMessage}` instead of `{`${label} is required`}`

[DEMO LINK](https://Augustwise.github.io/react_movies-list-add-form/)


Augustwise:
[DEMO LINK](https://Augustwise.github.io/react_movies-list-add-form/)

PR #3232, opened 2026-01-15
Solution

Milana-Danylivska:
[DEMO LINK](https://Milana-Danylivska.github.io/react_movies-list-add-form/)


de-vasta (changes requested):
Solid work here 👏🏼
That's nice that you [trimmed](https://github.com/mate-academy/react_movies-list-add-form/pull/3232/changes/a9aafe73675129374d76e5ef0ef095b5924d9c28#diff-6a86fc965e62c80cd45cffe4c8364ded78cab40471383b07990a0ab66476dfafR24) inputs before creating new `Movie` card, but it happens after `hasError` in `TextField.tsx` that checks whether `value` is valid. The side effect is, when I type only spaces for required fields, no error message shows up and nonetheless, send button is disabled. IMHO, that's confusing

I suggest, to add "trim logic" when the input isn't active (`onBlur`) or when we set value to `hasError`

de-vasta (approved):

PR #3233, opened 2026-01-15
✨ feat(movie): implement movie form functionality

de-vasta:
[Click Me](https://de-vasta.github.io/react_movies-list-add-form/)

### DIFF
--- a/src/components/NewMovie/NewMovie.tsx
+++ b/src/components/NewMovie/NewMovie.tsx
@@ -1,7 +1,7 @@
 import { useState } from 'react';
 import { TextField } from '../TextField';
 import { Movie } from '../../types/Movie';
-import { validateUrl } from '../../helpers/validateUrl';
+import { validateTitle, validateUrl } from '../../helpers/validations';
 
 interface Props {
   onAdd: (movie: Movie) => void;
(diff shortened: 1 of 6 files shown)


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,55 @@
...
+export const NewMovie = ({ onAdd }: Props) => {
   // Increase the count after successful form submission
   // to reset touched status of all the `Field`s
-  const [count] = useState(0);
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [descr, setDescr] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState<string>('');
```
dianakerekesha: It might be better to consolidate the state into a single formData object instead of five separate useState hooks. This would significantly reduce boilerplate by allowing a single handleChange function to manage all fields.

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,10 +1,55 @@
...
+
+  const isValidForSubmit = !!(title && imgUrl && imdbUrl && imdbId);
```
dianakerekesha: I would use .trim() during the check.

dianakerekesha (approved):
Good work! I have some minor concerns about the not-empty fields staying highlighted in red after losing focus, but if that was the intended design, I’m okay with it. Approved.

src/components/NewMovie/NewMovie.tsx
de-vasta: Yep, u got a point. All of them (or almost) related to one entity - `Movie`. Thx

PR #3243, opened 2026-01-16
Solution

mark-chep:


dmytro-halieba (approved):
Nice approach! Using a single state object for the movie data is much cleaner than managing multiple separate states. The handleChange function is elegant too!

PR #3332, opened 2026-03-31
Implement add movie and validation* 

Dianaeva:
[DEMO LINK](https://dianaeva.github.io/react_movies-list-add-form/)


polosanya (changes requested):
I still can create a movie with errors, the button should be disabled until you pass validation
Image

polosanya (approved):

PR #3362, opened 2026-05-03
Develop

OlehYuriev:
[DEMO LINK](https://OlehYuriev.github.io/react_movies-list-add-form/)


timurradkevic (approved):
Almost everything looks good👍 — the implementation is clean and meets most of the requirements. Just a small thing to improve: the submit button does not take URL validation into account, so it can be enabled even when the URLs are invalid. Overall, great job — only a minor tweak is needed.

PR #3363, opened 2026-05-05
implement solution

whomngmnt:
[DEMO LINK]  https://whomngmnt.github.io/react_movies-list-add-form/


src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,34 +1,110 @@
...
+  onAdd: (movie: Movie) => void;
+};
+export const NewMovie = ({ onAdd }: NewMovieProps) => {
+  const [count, setCount] = useState(0);
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [imgUrl, setImgUrl] = useState('');
+  const [imdbUrl, setImdbUrl] = useState('');
+  const [imdbId, setImdbId] = useState('');
+
```
781calmstate: 6 separate state that covers one form. you will have to track and update all three of them which creates bad dev experience and makes code to be harder to read.

I suggest to make it as single state that stores an object

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,34 +1,110 @@
...
+  const pattern =
+    // eslint-disable-next-line max-len
+    /^((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@,.\w_]*)#?(?:[,.!/\\\w]*))?)$/;
```
781calmstate: it's hardcoded regex, you can move it out of coponent

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,34 +1,110 @@
...
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+    setCount(currentCount => currentCount + 1);
+  };
+
+  const validateUrl = (label: string) => (value: string) => {
+    return pattern.test(value) ? '' : `${label} is not valid`;
+  };
```
781calmstate: it looks like a helper function, you can also move it out of component

src/components/NewMovie/NewMovie.tsx
```diff
@@ -1,34 +1,110 @@
...
+
+    setTitle('');
+    setDescription('');
+    setImgUrl('');
+    setImdbUrl('');
+    setImdbId('');
+    setCount(currentCount => currentCount + 1);
+  };
+
+  const validateUrl = (label: string) => (value: string) => {
```
781calmstate: this function isn't really validating anything, instead it returns another function which will perform the validation. For such function it's a common practice to name them something similar to `createUrlValidator` or smth similar, since this function creates another function

781calmstate (changes requested):

dothearts (approved):

PR #3368, opened 2026-05-07
Sol

sergeisvidlo4:
[DEMO](https://sergeisvidlo4.github.io/react_movies-list-add-form/)


CyborgNinjaHat (approved):
LGTM! Consider moving `pattern `outside the component to avoid recreating it on every render.

dothearts (approved):

PR #3375, opened 2026-05-07
done

vaynexx:


dothearts (changes requested):
Please, provide Demo link. 
Also, check lint and test pipelines, currently that crush. 

Many thanks!

PR #3380, opened 2026-05-09
finished task

Evgeniy45:
DEMO - https://Evgeniy45.github.io/react_movies-list-add-form/


timurradkevic (approved):
In general, the solution matches the requirements pretty well, but there are still a few issues that should be fixed:  `imgUrl` is not cleared after submit because `setImdbUrl('')` is duplicated. `isValid` returns a string instead of a boolean value.

dothearts (approved):

